PR #4, opened 2017-12-12
DB-1174: delete RPKI objects that are not referenced from manifests

erikrozendaal:
RPKI objects that are (no longer) referenced from current manifests
are deleted after at least 7 days of unreachability.

### DIFF
--- a/src/main/java/net/ripe/rpki/validator3/domain/RpkiObjectCleanupService.java
+++ b/src/main/java/net/ripe/rpki/validator3/domain/RpkiObjectCleanupService.java
@@ -1,0 +1,155 @@
+/**
+ * The BSD License
+ * <p>
+ * Copyright (c) 2010-2018 RIPE NCC
+ * All rights reserved.
+ * <p>
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * - Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * - Neither the name of the RIPE NCC nor the names of its contributors may be
+ * used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ * <p>
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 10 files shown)


src/main/java/net/ripe/rpki/validator3/domain/RpkiObjectCleanupService.java
```diff
@@ -0,0 +1,155 @@
...
+            RpkiObject.Type.MFT,
+            resourceCertificate.getSubjectKeyIdentifier()
+        );
+        maybeManifest.ifPresent(manifest -> {
+            markAndTraceObject(now, "manifest.mft", manifest);
+        });
+    }
+
+    private void markAndTraceObject(Instant now, String name, RpkiObject rpkiObject) {
+        if (now == rpkiObject.getLastMarkedReachableAt()) {
```
lolepezy: Is it a good idea to compare timestamps? 
Could there be inequality caused by inaccuracies?

src/main/java/net/ripe/rpki/validator3/domain/RpkiObjectCleanupService.java
```diff
@@ -0,0 +1,155 @@
...
+                break;
+            default:
+                break;
+        }
+    }
+
+    private void traceManifest(Instant now, String name, RpkiObject manifest) {
+        manifest.get(ManifestCms.class, name).ifPresent(manifestCms -> {
+            manifestCms.getFiles().forEach((entry, sha256) -> {
+                rpkiObjects.findBySha256(sha256).ifPresent(rpkiObject -> {
```
lolepezy: Shouldn't we better have rpkiObjects.findBySha256(listOfSha256s) method?
It's not that important, but as soon as we'll have a lot of object to mark, it could speed it up and avoid potential locks for other transactions (we had similar problem in current validator).

src/main/java/net/ripe/rpki/validator3/domain/RpkiObjectCleanupService.java
erikrozendaal: I actually use object instance identity here to see if an object has been marked during the current run. I'll add a comment, but it might be clearer (but slower) to have a special `visitedObjects` set.

src/main/java/net/ripe/rpki/validator3/domain/RpkiObjectCleanupService.java
erikrozendaal: This is definitely a `select 1+n` case. Not sure if querying a lists of objects will be that much faster (DB is local anyway). 

But sorting the hashes and then `SELECT FOR UPDATE` may reduce locking problems improve, so I'll add a `findObjectsInManifest` method.

PR #10, opened 2018-02-07
DB-1367: add separate files with internal and default properties

erikrozendaal:
Internal properties are configured in the
classpath:/application.properties file. Default properties are
specified in /etc/`app`/application-defaults.properties. They can be
overriden in /etc/`app`/application.properties.

The RPM still specifies /etc/`app`/application-defaults.properties as
a configuration file, which may be a problem when the
user (accidentally) modifies this file.

### DIFF
--- a/rpki-rtr-server/pom.xml
+++ b/rpki-rtr-server/pom.xml
@@ -346,6 +346,7 @@
                                     <location>src/main/resources/packaging/centos7/etc/rpki-rtr-server/</location>
                                     <includes>
                                         <include>application.properties</include>
+                                        <include>application-defaults.properties</include>
                                     </includes>
                                 </source>
                             </sources>
--- a/rpki-rtr-server/pom.xml
+++ b/rpki-rtr-server/pom.xml
@@ -312,8 +312,6 @@
                     <mappings>
                         <mapping>
                             <directory>/usr/lib/</directory>
-                            <username>rpki</username>
-                            <groupname>rpki</groupname>
                             <directoryIncluded>false</directoryIncluded>
                             <sources>
                                 <source>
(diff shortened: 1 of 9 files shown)


timbru:
Hi Erik,

Why do we store the defaults in a file in /etc/app.. rather than just packaging it? Is it your intent to have an explicit example of what could be overridden by users (regardless of whether it's a good idea to do so)? I see some merit in that, but then we should mark only the local file as a local configuration file in the RPM and make sure the defaults file gets overwritten when a new RPM is installed.

Personally I feel more for packaging the defaults in the jar, and having a config README file in /etc/app/.. that explains the user modifiable settings. This way we can be pretty sure that users don't accidentally change the wrong file. We should then of course still update this README when the RPM gets updated.

Or.. if packaging a README is cumbersome, we can create a config document here on GitHub in the Wiki and refer to that in "/etc/app/application.properties"

erikrozendaal:
Hi Tim,

The idea was that we'd have a self-documenting place to store user overridable defaults. Internal configuration properties are packaged in the jar from `src/main/resources/application.properties`. 

Maybe we can make `/etc/app/application-defaults.properties` read-only in the RPM package and add some documentation to put local configuration in `/etc/app/application.properties` only?

timbru:
Hi Erik,

I see now that  `/etc/app/application-defaults.properties` only includes the settings that are safe to update, even though users should rarely need them.

It works for me if you can make it readonly, add a warning to only modify the other file, and ensure that it does get updated when the RPM is upgraded.

lolepezy:
> The RPM still specifies /etc/app/application-defaults.properties as
> a configuration file, which may be a problem when the
> user (accidentally) modifies this file.

Why?
Shouldn't we have a comment in the "...default" file saying "don't modify this, modify another one if you feel like modifying " , remove "config" flag from it and let RPM overwrite it every time?

erikrozendaal:
Defaults properties file is now read-only and will be overwritten on upgrade.

PR #13, opened 2018-02-12
Spike ui

isvonja:


ui-validator/README.md
```diff
@@ -0,0 +1,28 @@
+# UiValidator
+
+This project was generated with [Angular CLI](https://github.com/angular/angular-cli) version 1.6.5.
+In package directory rpki-validator-3/ui-validator> run `npm install`. By default, npm install will install all modules listed as dependencies in package.json.
```
erikrozendaal: The README should specify that the node and npm installed by maven should be used, otherwise version conflicts will occur (errors on node_modules etc). What works for me is to modify your PATH to include the `./node` subdirectory first.

PR #14, opened 2018-02-13
Adding generic builds

timbru:
I had some spare cycles (or well created them I suppose) and I figured I wanted a generic build in addition to the RPMs for when I reach out to testers. So.. I did some set up:
* Moved some of the generic packaging stuff for the validator to generic folders for re-use
* Added simple start/stop/status scripts (inspired by validator-2.x)
* Added specific config for packaging, especially the validator needs other paths

It works locally on my Mac, but please double check things..


timbru:
As discussed in person - I will revise things a bit and re-use the existing start script, rather than duplicate and have something that looks like a real service script - but isn't..

rpki-validator/src/main/resources/packaging/generic/conf/application.properties
```diff
@@ -0,0 +1,94 @@
...
+#   garbage collection (slight speed improvements possible)
+jvm.memory.initial=512m       # -Xms jvm option -> initial memory claimed by the jvm
+jvm.memory.maximum=512m       # -Xmx jvm option -> maximum memory for the jvm
+
+#
+# The following directive is used to set where the rpki-validator keeps its
+# database.
+#
+# Note that the database cannot be shared by multiple instances
+spring.datasource.url=jdbc:h2:file:./db/rpki-validator.h2
```
erikrozendaal: `./data/db/rpki-validator.h2` ?

rpki-validator/src/main/resources/packaging/generic/conf/application.properties
```diff
@@ -0,0 +1,94 @@
...
+# The following directive is used to set where preconfigured TALs that ship with
+# the RPM can be found. This is only used when the rpki-validator-3 is first started.
+# If you wish to add (or remove) a TAL later, you will need to use the API for now.
+# A UI option may be added for this in future.
+rpki.validator.preconfigured.trust.anchors.directory=./preconfigured-tals
+
+#
+# The following directive is used to set where files retrieved from rsync based RPKI
+# repositories are kept.
+rpki.validator.rsync.local.storage.directory=./rsync
```
erikrozendaal: `./data/rsync` ?

erikrozendaal:
Looks good to me 👍

rpki-validator/src/main/resources/packaging/generic/conf/application.properties
timbru: I put all the directories directly under the unpacked project. Otherwise we would have ./data/data as well which I found odd.

PR #15, opened 2018-02-19
Ui db 1220 download

isvonja:


timbru:
Looks good to me from a functional point of view. I haven't looked at the code in enough detail to be able to say that this is the way. But it looks consistent and fine as far as I am concerned.

PR #29, opened 2018-04-13
Db 1557 search consistent

isvonja:


isvonja:
Just because of sql file V10__searchable_ignore_filter_prefix.sql I made this PR

PR #44, opened 2018-06-05
Add javax.xml.bind dependency to avoid problems with new JDKs

dacruz:


rpki-validator/pom.xml
```diff
@@ -87,6 +87,23 @@
             <version>1</version>
         </dependency>
 
+        <!-- add javax.xml dependency to avoid issues with new JDKs-->
+        <dependency>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>2.2.11</version>
```
dacruz: I'm not sure about the version numbers.

PR #59, opened 2019-01-08
Db 1884 add last modified to bgp preview

wibisono:


lolepezy:
As discussed, I would probably make Metadata parameterized on a type of this extra field. But as soon it can bring some JSON serialisation problems, we should probably leave it like it is.

PR #67, opened 2019-01-18
Background stat

wibisono:


rpki-validator/src/main/java/net/ripe/rpki/validator3/config/background/BackgroundJobInfo.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.TreeMap;
+
+@Component
+public class BackgroundJobInfo extends SchedulerListenerSupport {
+
+    private  TreeMap<String, Date> backgroundJobStats = new TreeMap<>();
+
+    @Override
+    public void jobScheduled(Trigger trigger) {
+        backgroundJobStats.put(((SimpleTriggerImpl) trigger).getJobName(), trigger.getStartTime());
```
lolepezy: Should we update the time in both jobScheduled and triggerFinalized?
Should it be the same time we are updating?

wibisono: Perhaps I was impatient while testing, so to not wait before all is finalised I also show scheduled, which can be confusing. I'll try to find a better way, maybe pick only one of them.

wibisono: I rewrite using JobListener instead of SchedulerListener, and listen to JobWasExecuted event, which is more precise and what we need.

PR #81, opened 2019-03-28
Avoiding unique constraint violation

wibisono:
Running this one on prepdev, startup from scratch seems reasonable, slightly more than 5 mins, and the last hour it has not yet produce any Unique Constraint Violation yet. *Crossing my finger*


rpki-validator/src/main/java/net/ripe/rpki/validator3/adapter/jpa/JPARpkiRepositories.java
```diff
@@ -156,7 +156,7 @@ public long countAll(RpkiRepository.Status optionalStatus, Long taId, boolean hi
     public Stream<RpkiRepository> findRsyncRepositories() {
         return stream(
             select()
-                .where(rpkiRepository.type.in(RpkiRepository.Type.RSYNC, RpkiRepository.Type.RSYNC_PREFETCH))
+                .where(rpkiRepository.type.in(RpkiRepository.Type.RSYNC))
```
wibisono: Actually my previous commit message explain this changes, somehow I mixed up commit messages.

But the intention here is to fetch only Rsync instead of Rsync or Prefetch. Keeping this as it is, might result in running two instance of repository validation for the same one. Because somewhere along the way, when CAT validation registering a repository it will modify type from Prefetch into Rsync. And these causes confusion, two jobs running the same repo and crying about unique constraint violation.

PR #85, opened 2019-05-07
Pilot branch allowing http

wibisono:
Only if we can live with this lax validation.


wibisono:
No need for this patch, pilot cert is fixed.

PR #101, opened 2019-10-16
Implement happy-eyeballs as a jetty http client's address resolver

omuravskiy:
This is a simplistic implementation of "happy-eyeballs" as described in RFC8305, but without sorting candidate IPs and caching results. Still falls back to v4 if v6 is broken, so could be released.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
@@ -1,0 +1,256 @@
+/*
+ * The BSD License
+ *
+ * Copyright (c) 2010-2018 RIPE NCC
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *   - Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *   - Redistributions in binary form must reproduce the above copyright notice,
+ *     this list of conditions and the following disclaimer in the documentation
+ *     and/or other materials provided with the distribution.
+ *   - Neither the name of the RIPE NCC nor the names of its contributors may be
+ *     used to endorse or promote products derived from this software without
+ *     specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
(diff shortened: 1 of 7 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
```diff
@@ -0,0 +1,256 @@
...
+                    }
+                    iterator.remove();
+                }
+            }
+        }
+        return null;
+    }
+
+    private static void awaitDnsResponses(ConcurrentLinkedQueue<InetAddress> resolvedAddressesV6,
+                                          ConcurrentLinkedQueue<InetAddress> resolvedAddressesV4) {
```
wibisono: Just trying to understand how this work. Before entering this await block we already spawn v4 and v6 lookup threads. Essentially this block tries to wait until we have any v6, with some deadline. 

What I am not sure I understand is that the flag to stop going. It is now within a block of if v4 is not empty. 

So it kinda assume that the lookup thread for v4 always returns a value (otherwise once entering the await, it won't stop). Do we have guarantee this loop won't stay when v4 lookup returns nothing?

omuravskiy: The logic according to the RFC should be:
- if the v6 address comes first, use it first
- if the v4 address comes first, wait 50ms more
- if v6 address comes within those 50 ms, use it first
- otherwise use v4 address first

omuravskiy: And, both lookups will return `ADDRESSES_TERMINATOR` at the end of their queues, so both `!isEmpty` conditions should become true at some point in time.

omuravskiy:
> I will look more carefully, but so far I am not super happy about using Threads directly. Could it be on-thread pool instead so that it's possible to start and cancel futures?

I thought using threads will be simpler.
What is the advantage of using Futures? (They would not have a result in our case.)
Using thread pools might have an advantage of not spawning new threads if they are available (but we only need 3), and a disadvantage of blocking if no threads are available in the pool.

lolepezy:
> I thought using threads will be simpler.
> What is the advantage of using Futures? (They would not have a result in our case.)
> Using thread pools might have an advantage of not spawning new threads if they are available (but we only need 3), and a disadvantage of blocking if no threads are available in the pool.

I don't know :)
It looks like this can be resolved by some concurrency primitives, but it doesn't come into my head which one, so I can't suggest anything else at the moment.

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
```diff
@@ -237,7 +237,7 @@ private boolean openConnectionFromQueue(Queue<InetAddress> queue) throws Interru
     }
 
     private static boolean queueExhausted(Queue<InetAddress> queue) {
-        return !queue.isEmpty() && queue.peek() == ADDRESSES_TERMINATOR;
+        return queue.peek() == ADDRESSES_TERMINATOR;
```
lolepezy: As a suggestion: do you want to have Queue<Optional<InetAddress>> and use 
queue.isEmpty() && !queue.peek().isPresent(), so you don't have to have a specific NULL objects?

omuravskiy: Yes, I think I like this idea…

lolepezy:
One more question: can we say anything about caching of the IPs in this case? I don't see anything in the Jetty code about caching. Shall we have it inside of the HappyEyeballsResolver?

Issue #40, opened 2018-06-04
Provide Debian package


job:
In addition to the CentOS package I think it would be good if Debian packages are also provided. To reduce workload, only Debian 9 needs to be supported at this moment.

dacruz:
Added to the backlog.

lolepezy:
Please try the Debian package here
https://ftp.ripe.net/tools/rpki/validator3/prod/deb/

Note that at the moment it is not a repository, so you will have to use something like 
apt install ./rpki-validator-XXX.deb, but we will start hosting the repository as the next step.

PR #114, opened 2019-11-05
Jdeb maven plugin

lolepezy:
This is to generate Debian packages together with RPMs (not finished yet, but for you to take a look). Open issue: https://github.com/RIPE-NCC/rpki-validator-3/issues/40


rpki-rtr-server/src/main/resources/packaging/deb/etc/systemd/system/rpki-rtr-server.service
```diff
@@ -0,0 +1,22 @@
+[Unit]
+Description=RPKI Validator
```
wibisono: Most probably my fault, RTR-Server should be here.

lolepezy: Done!

PR #120, opened 2019-12-05
Compare if new serial is higher than old serial

ties:
Warnings were logged while the serial was actually higher.


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/TrustAnchorValidationService.java
```diff
@@ -129,16 +129,17 @@ public void validate(long trustAnchorId) {
...
 
                     if (!validationResult.hasFailureForCurrentLocation()) {
-                        // validity time?
+                        // validate(..) is called multiple times for the same trust anchor certificate (e.g. when the
+                        // application restarts).
                         int comparedSerial = trustAnchor.getCertificate() == null ?
-                                1 : trustAnchor.getCertificate().getSerialNumber().compareTo(certificate.getSerialNumber());
+                                1 : newCertificate.getSerialNumber().compareTo(trustAnchor.getCertificate().getSerialNumber());
                         validationResult.warnIfTrue(comparedSerial < 0, ValidationString.VALIDATOR_REPOSITORY_OBJECT_IS_OLDER_THAN_PREVIOUS_OBJECT, trustAnchorCertificateURI.toASCIIString());
                         if (comparedSerial > 0) {
```
lolepezy: I thought we agreed on having it "if (comparedSerial != 0)", so if new serial is higher or lower that the old one, we proceed. If the new one is lower, we complain with a warning.

ties: Well spotted. Reverted this by accident when I git-reset the code that printed errors on duplicate serials.

PR #130, opened 2020-01-20
run unit tests with Gitlab actions (after #131)

### LINKED ISSUE #129: Tests fail on JDK 11 because Instant has nanoseconds
The test suite for this project fails on JDK 11, likely due to a change in Instant in JDK 9.

The root cause seems to be that `Instant`s now have nanosecond precision, while the encoder that encodes Instants as bytes only considers the milliseconds since the epoch.

https://github.com/RIPE-NCC/rpki-validator-3/blob/e77bfa8656422b4a62d1d08af62931e8e5d129b2/rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/encoding/custom/Coders.java#L51

While fixing this (by appending the nanoseconds) is easy, this will cause issues when the time since epoch no longer fits in a 64 bit number.

ties:
fixes #129

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -5,12 +5,15 @@ on: [push]
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        jdk: [1.8, 11]
 
     steps:
     - uses: actions/checkout@v1
-    - name: Set up JDK 1.8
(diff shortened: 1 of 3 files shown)


lolepezy:
Does it still make sense to have this one?

ties:
@lolepezy now it does – it was on top of the branch that we scrapped.

Note that the CI run would fail right now due to the JDK 11 issue.

PR #131, opened 2020-01-20
make project JDK 11 compatible by truncating instants

### LINKED ISSUE #129: Tests fail on JDK 11 because Instant has nanoseconds

ties:
fixes #129

### DIFF
--- a/rpki-rtr-server/pom.xml
+++ b/rpki-rtr-server/pom.xml
@@ -109,7 +109,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
-                <version>2.20.1</version>
+                <version>2.22.2</version>
             </plugin>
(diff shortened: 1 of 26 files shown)


ties:
@wibisono @lolepezy I switched this to the branch with the `InstantWithoutNanos` approach that was suggested.

Issue #48, opened 2018-09-19
API documentation/specification


nusenu:
It would be great to have parameter level documentation for the API.

https://www.ripe.net/ripe/mail/archives/routing-wg/2018-September/003615.html

devicenull:
This is important because the old API documentation isn't actually valid: https://www.ripe.net/support/documentation/developer-documentation/rpki-validator-api

nusenu:
dacruz:
> Yes, we need to update the documentation. 
> For the time being, you can check the swagger doc.
> It's not fully specified, but might help. 

As can be seen in the routing-wg email linked in the first
post of this issue, I opened this issue specifically because
the swagger doc does not include enough information.

PR #133, opened 2020-01-20
Improve api docs

ties:
* Improve the API documentation
  * Fix the `ValidationRun` API

Improves the situation for #48

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/ApiError.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/ApiError.java
@@ -42,9 +42,9 @@ public class ApiError {
     @ApiModelProperty(value = "(HTTP) status code", example = "500", required = true)
     String status;
     String code;
-    @ApiModelProperty(value = "Description of the error", example = "Internal Server Error", required = true)
+    @ApiModelProperty(value = "Reason for the error", example = "Internal Server Error", required = true)
     String title;
-    @ApiModelProperty(value = "Detailed description of error", required = false)
+    @ApiModelProperty(value = "More detailed description of the error", required = false)
     String detail;
     ApiErrorSource source;
 
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/ApiError.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/ApiError.java
@@ -30,6 +30,7 @@
 package net.ripe.rpki.validator3.api;
 
 import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
 import lombok.Builder;
 import lombok.Value;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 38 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/api/ApiError.java
```diff
@@ -38,9 +39,12 @@
 @Builder
 @ApiModel(value = "Error")
 public class ApiError {
+    @ApiModelProperty(value = "(HTTP) status code", example = "500", required = true)
     String status;
     String code;
+    @ApiModelProperty(value = "Description of the error", example = "Internal Server Error", required = true)
     String title;
+    @ApiModelProperty(value = "Detailed description of error", required = false)
```
lolepezy: It's either "the error" or "error" in both case? I still don't good intuition about it, but I think it would be more consistent.

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roaprefixassertions/RoaPrefixAssertionsController.java
```diff
@@ -109,7 +112,11 @@
         return ResponseEntity.ok(ApiResponse.data(toResource(roaPrefixAssertionsService.get(id))));
     }
 
-    @PostMapping(consumes = { Api.API_MIME_TYPE, "application/json" })
+    @ApiOperation(value = "Add whitelist entry", notes =
+            "By adding a whitelist entry you can manually authorise an ASN to originate a prefix in addition to the validated ROAs from the repository.\n" +
+            "Please note that whitelist entries may invalidate announcements for this prefix from other ASNs, just like ROAs. This may be intentional\n" +
+            "(you are whitelisting ASN A, ASN B is hijacking), or not (ASN B should also be authorised, or you made a mistake).")
```
lolepezy: "you are whitelisting ASN A if ASN B is hijacking"? Or that's exactly how you wanted it to be?

rpki-validator/src/main/java/net/ripe/rpki/validator3/jobs/JobExecutor.java
```diff
@@ -167,14 +167,14 @@ public void sequence(Job... newJobs) {
         private Job job;
     }
 
-    @EqualsAndHashCode
+    @EqualsAndHashCode(callSuper = true)
```
lolepezy: Oh! JobExecutor should be deleted at all, but somehow it leaked into the master branch, so don't bother changing it.

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/ApiError.java
ties: Well caught. I think it is "the error". I rephrased these two comments.

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roaprefixassertions/RoaPrefixAssertionsController.java
ties: This was not clear (but was taken from the web interface). I tried to clarify it.

ties: This was not clear (but was taken from the web interface). I tried to clarify it.

PR #135, opened 2020-01-22
Configure Single Page Application routes and test them

### LINKED ISSUE #134: curl of missing url on validator yields HTTP 406, this is confusing
When you try to retrieve a non-existing URL from the validator using curl, you get a HTTP 406 instead of the 404. As a user, this confuses me.

❯ curl http://localhost:8080/cache
```json
{
  "timestamp" : "2020-01-21 09:39:03",
  "status" : 406,
  "error" : "Not Acceptable",
  "message" : "Could not find acceptable representation",
  "path" : "/cache"
}

ties:
* Use addViewControllers over AngularController
  * Add tests for variations of test URLs

Fixes #134


ties:
A heads up: I got a NullPointerException from the StaticContentFixServlet on localhost just now. I think this is because the project was rebuilt after I started the validator but please check locally that it works.

(I tested most routes of the front-end locally, but four eyes is better than two)

lolepezy:
You need to also do mvn license:format and push the license headers

lolepezy:
Ah, sorry, my bad, I had local changes without license header.

ties:
> You need to also do mvn license:format and push the license headers

Good point though. Its not in my workflow to do that. Will check that the `verify` task checks that (and optionally update license to 2020).

ties:
During `verify` it checks that the license headers are present.

ties:
Tests pass in CI on my fork. Merged.

PR #139, opened 2020-01-31
Improve parallel execution

lolepezy:
In a nutshell it is to change from 
...
storage.readTx(tx -> retrieveManifestEntries(tx, manifest, manifestUri, temporary)))
.parallelStream()
...
to
... 
.parallelStream()
.map(entry -> getManifestEntry(manifestUri, entry))
...

In the existing implementation we do the following:
- gather all objects whose hashes are mentioned on the manifest into one big list
- do parallelStream() over this list

In the proposed implementation: 
- gather only the list of manifest entries
- do parallelStream() over this list and extract objects by hashes inside of the parallel chunks of work.

It has two advantages
1) more work is parallel (I did some very crude "visual benchmarking" and it's slightly faster)
2) we don't have to store all the objects pointed from the manifest in memory at any moment in time

Most of the trouble is fiddling with ValidationResult which is a mutable variable and has to be properly protected in synchronised blocks and using temporary ValidationResults.

Also, I saw some subtle differences in the numbers shown in the validator's UI, but these differences were always temporary and they always converge to the same numbers as the master branch. But I would strongly encourage you to run it locally and try to notice anything weird.


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -291,69 +290,108 @@ private boolean isValidationRunCompleted(ValidationResult validationResult) {
...
-                .filter(Optional::isPresent)
-                .map(Optional::get)
-                .collect(Collectors.toList())
+            manifest.getFiles().entrySet()
                 .parallelStream()
-                .map(childContext -> validateCertificateAuthority(trustAnchor, registeredRepositories, childContext, temporary))
-                .forEachOrdered(validatedObjects::addAll);
+                .map(entry -> getManifestEntry(manifestUri, entry))
+                .filter(t -> t.v2().isPresent())
+                .map(t -> t.map2(Optional::get))
```
wibisono: Instead of map filter optional, I think flatmap and returning stream/empty stream is nicer, see commit on separate branch:
https://github.com/RIPE-NCC/rpki-validator-3/commit/9dd3321d80c21e2a13f4941d976a0807e388fe72

If it make sense you can merge it here, not worth separate PR.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -291,69 +290,108 @@ private boolean isValidationRunCompleted(ValidationResult validationResult) {
...
+                        validatedObjects.addAll(t.v1());
+                    }
+                    synchronized (temporary) {
+                        temporary.addAll(t.v2());
+                    }
+                });
 
         } catch (Exception e) {
-            synchronized (this) {
+            synchronized (validationResult) {
```
wibisono: Idea is unhappy synchronizing on method parameter.

ties: More synchronization is needed (line 231, probably at calls where it is passed in as an argument as well).

I do not think locking on the argument is feasible.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -291,69 +290,108 @@ private boolean isValidationRunCompleted(ValidationResult validationResult) {
...
                 validationResult.addAll(temporary);
             }
         }
 
         return validatedObjects;
     }
 
+    private Tuple3<URI, Optional<RpkiObject>, ValidationResult> getManifestEntry(URI manifestUri, Map.Entry<String, byte[]> entry) {
```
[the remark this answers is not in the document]
lolepezy: It should be less ugly now

PR #143, opened 2020-02-14
Log on the first successful cache update after a failure

ties:


ties:
new output:
Screenshot 2020-02-14 at 17 39 46

PR #147, opened 2020-02-27
Include the manifest URL in the log message

ties:
Fixes a missing parameter, causing messages like `Manifest https://rpki.cnnic.cn/rrdp/notify.xml for this certificate is not available at repository {1}.`.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
@@ -271,7 +271,7 @@ public class CertificateTreeValidationService {
                 if (rpkiRepository.getStatus() == RpkiRepository.Status.FAILED) {
                     temporary.error(ValidationString.VALIDATOR_NO_MANIFEST_REPOSITORY_FAILED, rpkiRepository.getLocationUri());
                 } else {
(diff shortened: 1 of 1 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -271,7 +271,7 @@ private void validateCertificateAuthority(final TrustAnchor trustAnchor,
                 if (rpkiRepository.getStatus() == RpkiRepository.Status.FAILED) {
                     temporary.error(ValidationString.VALIDATOR_NO_MANIFEST_REPOSITORY_FAILED, rpkiRepository.getLocationUri());
                 } else {
-                    temporary.error(ValidationString.VALIDATOR_NO_LOCAL_MANIFEST_NO_MANIFEST_IN_REPOSITORY, rpkiRepository.getLocationUri());
+                    temporary.error(ValidationString.VALIDATOR_NO_LOCAL_MANIFEST_NO_MANIFEST_IN_REPOSITORY, manifestUri.toString(), rpkiRepository.getLocationUri());
```
lolepezy: I think it requires change in rpki-commons as well. Is it already there?

ties: It already is in rpki-commons, that is how I found out:
Screenshot 2020-02-28 at 15 37 53

PR #150, opened 2020-03-03
We don't really need core utils.

wibisono:
Managed to install this in xenial (vagrant).


lolepezy:
> curl - for the upload-ta script (at least as recommended)
Indeed! That's a good catch.

> * which
Isn't "which" a built-in shell command?

ties:
> > * which
> >   Isn't "which" a built-in shell command?

I tested it in a minimal environment. It appears to be included there (as well as `/bin/sh`).

wibisono:
Curl added, debianutils seems okay to leave out.

ties:
@wibisono I think my proposed change wasn't a good idea. Merged 👍

PR #151, opened 2020-03-05
Upgrade to spring boot 2.2 and add prometheus endpoint

ties:
I'll run this for a while in my dev environment to make sure it works and resource usage is similar to last friday's branch.

Confirmed that test suite passes locally.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/metrics/TrustAnchorMetricsService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/metrics/TrustAnchorMetricsService.java
@@ -43,14 +43,9 @@ public class TrustAnchorMetricsService {
             return;
         }
 
-        if (cetrificateTreeValidationMetrics.containsKey(uri)) {
-            cetrificateTreeValidationMetrics.get(uri).update(vr, durationMs);
-        } else {
-            final CertificateTreeValidationMetrics oldMetrics = cetrificateTreeValidationMetrics.putIfAbsent(uri, new CertificateTreeValidationMetrics(ta, vr, durationMs));
-            if (oldMetrics != null) {
-                oldMetrics.update(vr, durationMs);
-            }
-        }
+        cetrificateTreeValidationMetrics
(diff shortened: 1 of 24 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/metrics/TrustAnchorMetricsService.java
```diff
@@ -0,0 +1,143 @@
...
+    private ConcurrentHashMap<String, CertificateTreeValidationMetrics> cetrificateTreeValidationMetrics = new ConcurrentHashMap<>();
+
+    public void update(TrustAnchor ta, CertificateTreeValidationRun vr, long durationMs) {
+        final String uri = ta.getLocations().size() > 0 ? ta.getLocations().get(0) : null;
+        if (uri == null) {
+            log.error("Trust anchor {} does not have a location.", ta.getSubjectPublicKeyInfo());
+            return;
+        }
+
+        if (cetrificateTreeValidationMetrics.containsKey(uri)) {
```
lolepezy: Strictly speaking, ConcurrentHashMap doesn't protect from race conditions between calls of containsKey and get in this case. 

This should work atomically 
cetrificateTreeValidationMetrics
   .putIfAbsent(uri, new CertificateTreeValidationMetrics(ta, vr, durationMs))
   .update(vr, durationMs);

lolepezy: Or usual map and synchronised block.

ties: I was doing the initial update from the constructor. I made this slightly cleaner.

ties: It did not matter if the object was constructed twice. I cleaned this by moving the initial update from the constructor and doing that explicitly.

Issue #39, opened 2018-06-04
Provide more detail in export.json output


job:
Currently we have this output:

```{
  "roas" : [ {
    "asn" : "AS0",
    "prefix" : "2001:7fa:0:3::/64",
    "maxLength" : 128,
    "ta" : "APNIC RPKI Root"
  },
  ....
```
It would be great if we can get some more detail, like validity period, serial, etc:
```{
  "roas" : [ {
    "asn" : "AS0",
    "prefix" : "2001:7fa:0:3::/64",
    "maxLength" : 128,
    "ta" : "APNIC RPKI Root",
    "validity_period": "2018-06-04T05:53:04.000Z - 2018-06-11T05:58:04.000Z",
    "serial_nr": "4398793361",
  },
  ....
```

dacruz:
Added to the backlog.

lolepezy:
We have this end-point since recently, 
https://rpki-validator.ripe.net/api/export-only-roas.json
Added to the version 3.1-2020.05.08.09.26.49 that we just released.

PR #152, opened 2020-03-05
Add more details on export #39

wibisono:
Not sure if we wanted to do this, but have a look, maybe worth it.

See: https://github.com/RIPE-NCC/rpki-validator-3/issues/39

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roas/ExportsController.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roas/ExportsController.java
@@ -55,8 +55,10 @@ import static net.ripe.rpki.validator3.api.ModelPropertyDescriptions.SERIAL_NUMB
 import static net.ripe.rpki.validator3.api.ModelPropertyDescriptions.SERIAL_NUMBER_EXAMPLE;
 import static net.ripe.rpki.validator3.api.ModelPropertyDescriptions.TRUST_ANCHOR;
 import static net.ripe.rpki.validator3.api.ModelPropertyDescriptions.TRUST_ANCHOR_EXAMPLE;
-import static net.ripe.rpki.validator3.api.ModelPropertyDescriptions.VALIDITY_PERIOD;
-import static net.ripe.rpki.validator3.api.ModelPropertyDescriptions.VALIDITY_PERIOD_EXAMPLE;
+import static net.ripe.rpki.validator3.api.ModelPropertyDescriptions.VALIDITY_PERIOD_NOT_AFTER;
+import static net.ripe.rpki.validator3.api.ModelPropertyDescriptions.VALIDITY_PERIOD_NOT_AFTER_EXAMPLE;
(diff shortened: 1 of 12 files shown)


lolepezy:
It's a breaking change of, probably, the most widely used end-point in the API.
While JSON parser will likely be ok with it, we can't guarantee the CSV parsers 
won't break, for example.

So it's a nice to have, but can you move it to another end-point? export-extended or something.

wibisono:
Added separate button and separate export-extended.json end point.

rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/encoding/custom/RoaPrefixCoder.java
```diff
@@ -60,8 +67,11 @@ public RoaPrefix fromBytes(byte[] bytes) {
         long asn = Coders.toLong(content.get(ASN_TAG));
         byte[] maxLen = content.get(MAX_LEN_TAG);
         Integer maximumLength = maxLen != null ? Coders.toInt(maxLen) : null;
+        long notBefore = Coders.toLong(content.get(VALIDITY_BEFORE));
```
ties: How will this work when migrating from an old db? I think that `Coders.toLong` may throw an IllegalArgumentException.

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roas/ExportsController.java
```diff
@@ -133,6 +142,32 @@ public void exportCsv(HttpServletResponse response) throws IOException {
...
+
+        Stream<JsonRoaPrefixExtended> validatedPrefixes = validatedRpkiObjects
+                .findCurrentlyValidatedRoaPrefixes()
+                .getObjects()
+                .map(r -> new JsonRoaPrefixExtended(
+                        String.valueOf(r.getAsn()),
+                        r.getPrefix().toString(),
+                        r.getEffectiveLength(),
+                        r.getTrustAnchor().getName(),
+                        Instant.ofEpochMilli(r.getNotBefore()) + " - " + Instant.ofEpochMilli(r.getNotAfter()),
```
ties: I think that as a user I'd rather consume this in two separate fields.

wibisono: Separated them.

rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/encoding/custom/RoaPrefixCoder.java
wibisono: Added a guard checking availability of these fields. Now still testing old db/new version and how it behaves.

wibisono: Without DB clean up, these additional validity fields will not be updated, validator were not aware they are modified since hash is the same, so no refetch/update db. But it does not crash :)

ties:
Great! I think you can merge it.

I'll update my test environment to monitor resource usage.

PR #153, opened 2020-03-11
Add http status code as prometheus metrics

ties:
Metrics for HTTP status codes (and error messages). Example metrics:
```
# HELP http_response_status_total HTTP request result (per server, per status)
# TYPE http_response_status_total counter
http_response_status_total{status="200",url="https://rrdp.apnic.net/",} 1.0
http_response_status_total{status="200",url="https://www.ris.ripe.net/",} 2.0
http_response_status_total{status="javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure",url="https://rrdp.ripe.net/",} 1.0
```

This should help me debug the amount of SSLHandshakeExceptions…

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpRisDownloader.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpRisDownloader.java
@@ -33,6 +33,7 @@ import lombok.extern.slf4j.Slf4j;
 import net.ripe.ipresource.Asn;
 import net.ripe.ipresource.IpRange;
 import net.ripe.ipresource.UniqueIpResource;
+import net.ripe.rpki.validator3.domain.metrics.HttpClientMetricsService;
 import net.ripe.rpki.validator3.util.Http;
 import org.eclipse.jetty.client.HttpClient;
 import org.eclipse.jetty.client.api.Request;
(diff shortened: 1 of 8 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpRisDownloader.java
```diff
@@ -86,6 +89,9 @@ public void preDestory() throws Exception {
 
     public BgpRisDump fetch(@NotNull BgpRisDump dump) {
         log.info("attempting to download new BGP RIS preview dump from {}", dump.url);
+        long before = System.currentTimeMillis();
```
lolepezy: Just in case: there's net.ripe.rpki.validator3.util.Time class, maybe it can make the code simpler here.

rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/HttpRrdpClient.java
```diff
@@ -69,16 +74,21 @@ public void postConstruct() throws Exception {
 
     @Override
     public <T> T readStream(final String uri, Function<InputStream, T> reader) {
+        long before = System.currentTimeMillis();
```
lolepezy: Same here.

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpRisDownloader.java
ties: It's a nice utility (and I had not considered using it here), but I think that this is not convenient here (since http statuses are communicated through exceptions).

ties:
Response time is measured without considering the http status code. I don't see value in splitting these up by status code (for now).

PR #156, opened 2020-03-18
Unwrap sub-type of exception first

ties:
Cause was empty for `HttpStatusException`. Now we handle `Http.Failure`'s without a cause properly.

This may cause some unique log messages (and transitively unique metrics in prometheus). I'll keep an eye on that.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/metrics/HttpClientMetricsService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/metrics/HttpClientMetricsService.java
@@ -83,7 +83,7 @@ public class HttpClientMetricsService {
                 return rootCause.toString();
             }
         }
-        return cause.toString();
+        return cause.getClass().getName();
     }
 
     public static class HttpStatusMetric {
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/metrics/HttpClientMetricsService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/metrics/HttpClientMetricsService.java
@@ -82,9 +82,8 @@ public class HttpClientMetricsService {
                 }
                 return rootCause.toString();
             }
-            return cause.toString();
         }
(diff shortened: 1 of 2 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/metrics/HttpClientMetricsService.java
```diff
@@ -69,17 +69,20 @@ public void update(String uri, String statusDescription, long durationMs) {
...
+                    return rootCause.getClass().getName();
+                }
+                return rootCause.toString();
             }
-            return rootCause.toString();
-        } else if (cause instanceof Http.HttpStatusException) {
-            return String.valueOf(((Http.HttpStatusException)cause).getCode());
+            return cause.toString();
         }
         return "exception";
```
lolepezy: Not sure 'return "exception"' is very useful. 
Maybe String.valueOf(cause)?

ties: Some of the exceptions have string representations that are unique for each instance. I could try `cause.toString()` and monitor whether the cardinality of the status field explodes.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/metrics/HttpClientMetricsService.java
```diff
@@ -69,19 +69,21 @@ public void update(String uri, String statusDescription, long durationMs) {
...
+                    return rootCause.getClass().getName();
+                }
+                return rootCause.toString();
             }
-            return rootCause.toString();
-        } else if (cause instanceof Http.HttpStatusException) {
-            return String.valueOf(((Http.HttpStatusException)cause).getCode());
         }
-        return "exception";
+        return cause.toString();
```
lolepezy: As a compromise, cause.getClass().toString(), but I don't really expect a lot of different exceptions.

PR #163, opened 2020-03-30
Updated docker container

ties:
Adds a build for a docker container to the repo.

I decided to go with a build from GitLab CI because this ensures it matches the lifecycle of our other artifacts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,57 @@
+# First build step: check hash and unpack arcive
+#
+# The hash ensures that args change even if file name is equal, otherwise docker
+# would cach an image when GENERIC_BUILD_ARCHIVE changes even though the file
+# changes.
+FROM openjdk:8-jre-alpine as intermediate
+ARG GENERIC_BUILD_ARCHIVE
+ARG GENERIC_BUILD_SHA256
+
+COPY ${GENERIC_BUILD_ARCHIVE} /tmp/
+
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+ENV JVM_XMX=""
+# Used by `rpki-validator-3.sh`
+ENV CONFIG_DIR="/config"
+
+COPY --from=intermediate /opt/rpki-validator-3 /opt/rpki-validator-3
+WORKDIR /opt/rpki-validator-3
+
+RUN apt-get update && apt-get install --no-install-recommends --yes rsync \
+    # Clean apt cache
+    && rm -rf /var/lib/apt/lists/* \
```
wibisono: Wondering about this single RUN with multiple &&, and it seems the answer to which one we should follow changes over time :D 

https://stackoverflow.com/questions/39223249/multiple-run-vs-single-chained-run-in-dockerfile-what-is-better

ties: I'm using multistage builds. Otherwise the archive would be added to the disk (and add a layer) and later be unpacked. However, the file that you `ADD`/`COPY` never can be removed.

In the second step I need to adjust some parts of the config file. It is not an expensive layer that it would add but the layer is also not reusable if the archive changes. That is why I went to multiline-style (I'm used to it).

PR #164, opened 2020-03-31
Fix changes due to new dependencies from snapshot of rpki-commons

ties:


rpki-validator/src/main/java/net/ripe/rpki/validator3/util/TrustAnchorLocator.java
```diff
@@ -172,14 +172,6 @@ public String getPublicKeyInfo() {
         return prefetchUris;
     }
 
-    @Override
-    public String toString() {
```
wibisono: I hope this is only for logging (nobody consume or use it for other purpose). Unless lombok.ToString produces same output?

ties: I think it is only used for debugging and the lombok to string should produce similar output.

ties: I had a  breakpoint there and did not hit it.

PR #168, opened 2020-04-06
Revert jetty to avoid older ssl handshake issue.

wibisono:


ties:
This ran fine in my environment.

But if we're hitting issues due to behavioral blacklisting of clients this will not help.

PR #171, opened 2020-04-07
Improve the wording of listen address config property

### LINKED ISSUE #166: Please fix comment in conf/application.properties
In conf/application.properties it is described that "By default the validator will listen on localhost ONLY" which is a good thing.
However it also states "You can change this behavior by uncommenting the following line".

This line however is already uncommented and should be commented out if one wants to change the behavior.

Hence I would like to propose to change:
"You can change this behavior by uncommenting the following line"
to
"You can change this behavior by commenting out the following line"

Thanks

ties:
If accepted this should fix #166.

### DIFF
--- a/rpki-rtr-server/src/main/resources/application.properties
+++ b/rpki-rtr-server/src/main/resources/application.properties
@@ -47,21 +47,24 @@ server.port=8081
 rtr.server.port=8323
 
 #
-# By default rpki-rtr-server will listen on localhost ONLY for RTR connections
-#
-rtr.server.address=localhost
-
-
-# By default the web-interface of rpki-rtr-server will listen on localhost ONLY
-#
-# You can change this behavior by uncommenting the following line in
-# which case the validator will listen to on all interfaces, or by
-# specifying a different IP address
+# You can specify the address the rpki-rtr-server listens on by setting the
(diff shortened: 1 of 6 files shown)


rpki-rtr-server/src/main/resources/application.properties
```diff
@@ -47,22 +47,25 @@ server.port=8081
 rtr.server.port=8323
 
 #
-# By default rpki-rtr-server will listen on localhost ONLY for RTR connections
+# You can specify the address the rpki-rtr-server listens on by setting the
+# rtr.server.address property. If no server.address is set rtr-server will listen
+# on all interfaces (on IPv4 and IPv6).
+# If you want to explicitly listen to all (IPv4) addresses use 0.0.0.0
 #
+# When you do we recommend that you limit access to this service in
```
lolepezy, line 55: "When you do we recommend that you limit access to this service in your firewall"
When you do what?

ties: This is a verbatim copy of somewhere else in the file.

Proposal:
```suggestion
# When your rpki-rtr-server listens on all interfaces we recommend that you
# limit access to this service using a firewall.
```

PR #175, opened 2020-04-15
Correct format specifier for seconds

ties:


ui-validator/src/app/monitoring-ta/validation-details/validation-details.component.html
```diff
@@ -14,7 +14,7 @@ <h3><strong>{{'ValidationDetails.TITLE' | translate}}</strong></h3>
     </thead>
     <tbody>
     <tr *ngFor='let validationCheck of validationChecks'>
-      <td>{{validationCheck.createdAt | date: 'yyyy-MM-dd HH:MM:SS' }}</td>
+      <td>{{validationCheck.createdAt | date: 'yyyy-MM-dd HH:MM:ss' }}</td>
```
lolepezy, line 17: what does it do?

ties: I made a mistake that I did not see in testing. I just spotted it because a time ended in an impossible number of seconds (`2020-04-15 14:04:69`).

It was using the specifier for fractional seconds ([fractional seconds, two digits](https://angular.io/api/common/DatePipe)) instead of seconds (`ss`).

PR #177, opened 2020-04-21
Bump spring boot version for new Jackson version

ties:
No security implications because Hikari is not on classpath and we do'nt deserialize polymorphic types (let alone polymorphic types with open bounds).

This update keeps a dark cockpit in whitesource.

### DIFF
--- a/rpki-parent/pom.xml
+++ b/rpki-parent/pom.xml
@@ -20,7 +20,8 @@
 
         <rpki-commons.version>1.8.0</rpki-commons.version>
 
-        <micrometer.version>1.3.5</micrometer.version>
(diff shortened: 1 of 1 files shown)


ties:
Test suite is flaky right now (fails on master) due to an integration test using a broken rrdp depository.

ties:
@wibisono can you take a look at the test that fails?

lolepezy:
Do we need to have this broken action? I can't understand what is it trying to do, to be honest.

ties:
The action should do a basic compile/test. However the infrastructure seems to have had issues (https://www.githubstatus.com) and the tests were flaky.

I'll remove it.

Issue #185, opened 2020-05-02
SLURM (Whitelisting/Blacklisting) is not honored for the /api/export.json endpoint


markusju:
> The output will be in CSV or JSON format and consist of all validated ROAs, minus your ignore filter entries, plus your whitelist entries.

I just noticed that contrary to the statements found in the UI, the data returned from `/api/export.json` endpoint does not honor white- and/or blacklisting entries. 

SLURM is correctly processed when retrieving the data from the `/api/objects/validated` endpoint.

If this is intended behavior, I would suggest to fix the description in the frontend. If not one could clarify the functionality of the different endpoints and adjust their behavior.

Thanks! :)

markusju:
As far as I can tell from the code, white and blacklisting is processed explicitly (and exclusively) for the `/api/objects/validated` endpoint.

This sucks, as I can't use the data produced by this endpoint for a rtr-server because the schema is not compliant with the "de-facto" standard.

Any plans to fix this?

lolepezy:
Hi Markus,

It's a good catch and we'll going to fix it. I am inclined to add it as a separate end-point, not to break the API.

lolepezy:
Fixed in the version 3.1-2020.05.08.09.26.49 that we just released.

markusju:
Thanks guys, much appreciated! :)

PR #187, opened 2020-05-07
Use SLURM rules for export.{csv/json}

lolepezy:
Bug reported here
https://github.com/RIPE-NCC/rpki-validator-3/issues/185

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roas/ExportsController.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roas/ExportsController.java
@@ -36,6 +36,9 @@ import io.swagger.annotations.ApiOperation;
 import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
 import net.ripe.rpki.validator3.api.PublicApiCall;
+import net.ripe.rpki.validator3.api.ignorefilters.IgnoreFilterService;
+import net.ripe.rpki.validator3.api.roaprefixassertions.RoaPrefixAssertionsService;
+import net.ripe.rpki.validator3.domain.IgnoreFiltersPredicate;
 import net.ripe.rpki.validator3.domain.validation.ValidatedRpkiObjects;
 import net.ripe.rpki.validator3.storage.Storage;
 import net.ripe.rpki.validator3.storage.stores.Settings;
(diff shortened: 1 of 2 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roas/ExportsController.java
```diff
@@ -123,74 +121,114 @@ public void exportCsv(HttpServletResponse response) throws IOException {
...
     }
 
+    /**
+     * Get the stream of VRPs based on ROAs - whitelist + SLURM assertions.
+     */
+    private Stream<VRP> vrpStream() {
+        Stream<VRP> validatedPrefixes = validatedRpkiObjects
+            .findCurrentlyValidatedRoaPrefixes()
+            .getObjects()
+            .filter(new IgnoreFiltersPredicate(ignoreFilters.all()).negate())
```
wibisono, line 198: Somehow we need to mark PR with refactoring where to focus on most relevant changes :P 

Here is what changed and makes difference on the whole story :)

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roas/ExportsController.java
```diff
@@ -123,74 +121,114 @@ public void exportCsv(HttpServletResponse response) throws IOException {
...
+        final Stream<VRP> assertions = roaPrefixAssertions
+            .all()
+            .map(assertion -> new VRP(
+                assertion.getAsn().toString(),
+                assertion.getPrefix().toString(),
+                assertion.getMaxPrefixLength() != null ? assertion.getMaxPrefixLength() : assertion.getPrefix().getPrefixLength(),
+                "SLURM"
+            ));
+
+        return Stream.concat(validatedPrefixes, assertions).distinct();
```
ties, line 215: VRPs with an differing `ta` will not be seen as distinct (so SLURM versus from any of the trust anchors will both be there)

ties:
Rebased: Should remove the failing checks since they are no longer in master.

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roas/ExportsController.java
lolepezy: Yeah :)
Just had a fit of boyscouting.

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/roas/ExportsController.java
lolepezy: I guess the point was to just return all VRPs regardless of their priority or origin, so conflict resolution and prioritising is up to the consumer.

lolepezy:
Fixed in the version 3.1-2020.05.08.09.26.49 that we just released.

PR #192, opened 2020-05-15
Fix configuration property priority

wibisono:
Defaults should override only when key is not defined in application.properties.
Also use jvm.memory instead of jvm.mem


rpki-validator/src/main/resources/packaging/generic/bin/rpki-validator-3.sh
```diff
@@ -64,10 +64,13 @@ CONFIG_DEFAULT_FILE="${CONFIG_DIR}/application-defaults.properties"
 
 function parse_config_line {
     local CONFIG_KEY="$1"
```
wibisono, line 66: Here is the main change, order of looking up standard up first and fall back to default.

PR #193, opened 2020-05-15
Throttle tree revalidation

lolepezy:
This is to prevent CPU overuse by the constant re-validation after every repository update. As we have more and more small delegated repositories, it's getting more and more expensive. The logic in the PR is to "execute not more often than the minimal repository update interval".


rpki-validator/src/main/java/net/ripe/rpki/validator3/background/ValidationScheduler.java
```diff
@@ -95,28 +123,12 @@ public boolean scheduledTrustAnchor(TrustAnchor trustAnchor) {
...
         try {
-            return scheduler.checkExists(TrustAnchorValidationJob.getJobKey(trustAnchor)) &&
-                    scheduler.checkExists(CertificateTreeValidationJob.getJobKey(trustAnchor));
+            return scheduler.checkExists(TrustAnchorValidationJob.getJobKey(trustAnchor));
         } catch (SchedulerException e) {
             return false;
         }
     }
 
-    public void removeTrustAnchor(TrustAnchor trustAnchor) {
```
wibisono, line 105: Unused/dead code?

lolepezy: Yeah, it wasn't actually used, it's some boyscouting.

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
```diff
@@ -0,0 +1,119 @@
...
+                    }
+                }
+            } else {
+                actionMap.put(key, new Action(now, true));
+                scheduledExecutor.execute(wrappedRunnable);
+            }
+        }
+    }
+
+    @Data
```
ties, line 108: @Value iff we're using this as immutable objects?

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/ValidationScheduler.java
```diff
@@ -42,28 +45,54 @@
...
                                @Value("${rpki.validator.rsync.repository.download.interval:PT10M}") String rsyncRepositoryDownloadInterval,
-                               @Value("${rpki.validator.rrdp.repository.download.interval:PT2M}") String rrpdRepositoryDownloadInterval) {
+                               @Value("${rpki.validator.rrdp.repository.download.interval:PT2M}") String rrpdRepositoryDownloadInterval,
+                               @Lazy CertificateTreeValidationService validationService) {
         this.scheduler = scheduler;
         this.rsyncRepositoryDownloadInterval = Duration.parse(rsyncRepositoryDownloadInterval);
         this.rrpdRepositoryDownloadInterval = Duration.parse(rrpdRepositoryDownloadInterval);
+        this.validationService = validationService;
+
+        // Allow tre re-validation as often as minimum repository fetch interval,
```
ties, line 88:
```suggestion
        // Allow tree re-validation as often as minimum repository fetch interval,
```

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/ValidationScheduler.java
```diff
@@ -42,28 +45,54 @@
...
+
+    private final Throttled<String> throttledTreeValidation;
+
     private boolean enabled = true;
 
+    private final CertificateTreeValidationService validationService;
+
     @Autowired
     public ValidationScheduler(Scheduler scheduler,
                                @Value("${rpki.validator.rsync.repository.download.interval:PT10M}") String rsyncRepositoryDownloadInterval,
```
ties, line 80: Not a change, but: `@Value(...) Duration argName` would be possible iirc

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
```diff
@@ -0,0 +1,119 @@
...
+public class Throttled<Key> {
+
+    private final long minInterval;
+
+    private final Map<Key, Action> actionMap = new HashMap<>();
+
+    private final ScheduledExecutorService scheduledExecutor =
+        Executors.newScheduledThreadPool(Runtime.getRuntime().availableProcessors());
+
+    public Throttled(long minInterval) {
```
wibisono, line 69: Some comments/variable rename to indicate units. Which of course couple of minutes later reading the implementation we can see it's seconds, but worth to safe future reader's time.

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/ValidationScheduler.java
lolepezy: Makes sense, but I would do it separately as we have multiple places with these intervals.

rpki-validator/src/test/java/net/ripe/rpki/validator3/background/ThrottledTest.java
```diff
@@ -29,47 +29,46 @@
...
+        assertEquals(1, counter.get());
 
-    @Override
-    public void execute(JobExecutionContext context) throws JobExecutionException {
-        validationService.validate(trustAnchorId);
+        for (int i = 0; i < 10; i++) {
+            throttled.trigger("x", counter::incrementAndGet);
+        }
+        Thread.sleep(1100);
+        assertEquals(1, counter.get());
```
wibisono, line 52: If I understand correctly, the first of this 10 additional trigger will be delayed for around 10 secs, and the rest will be rejected. So this will also holds true immediately after.
```
        Thread.sleep(10000);
        assertEquals(2, counter.get());
```

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
lolepezy: Yeah, that's a good idea.

rpki-validator/src/test/java/net/ripe/rpki/validator3/background/ThrottledTest.java
lolepezy: I fixed the test so that it also checks for the case of "immediately trigger after".

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
lolepezy: Yeap!

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
```diff
@@ -0,0 +1,119 @@
...
+                synchronized (actionMap) {
+                    actionMap.put(key, new Action(after, false));
+                }
+            }
+        };
+
+        final Instant now = Instant.now();
+        synchronized (actionMap) {
+            final Action action = actionMap.get(key);
+            if (action != null) {
```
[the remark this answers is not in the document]
lolepezy: Yes, but ConcurrentHashMap would require using updateIfAbsent and it would 
look quite strange to do side-effect inside of an atomic action.

Issue #194, opened 2020-05-18
SLURM entries with a single parameter cause a Java exception


jlsnt:
Hello,

Upon creating a SLURM validation output filter with only one parameter, i.e. specifying only a prefix **OR** an ASN, a Java exception is thrown and the filter is not taken into account by the validator.

RFC 8416 allows such prefix filters in § 3.3.1.

May you please have a look at this ?
Thank you :)

```
$ java -version
openjdk version "11.0.7" 2020-04-14
OpenJDK Runtime Environment (build 11.0.7+10-post-Ubuntu-2ubuntu218.04)
OpenJDK 64-Bit Server VM (build 11.0.7+10-post-Ubuntu-2ubuntu218.04, mixed mode, sharing)
```

```
2020-05-18 15:05:32,144          ERROR   XNIO-1 task-46                          i.u.request             UT005023: Exception handling request to /api/ignore-filters
org.springframework.web.util.NestedServletException: Request processing failed; nested exception is java.lang.NullPointerException
        at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
        at org.springframework.web.servlet.FrameworkServlet.doPost(FrameworkServlet.java:909)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:523)
        at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:883)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:590)
        at io.undertow.servlet.handlers.ServletHandler.handleRequest(ServletHandler.java:74)
        at io.undertow.servlet.handlers.FilterHandler$FilterChainImpl.doFilter(FilterHandler.java:129)
        at org.springframework.web.filter.RequestContextFilter.doFilterInternal(RequestContextFilter.java:100)
        at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
        at io.undertow.servlet.core.ManagedFilter.doFilter(ManagedFilter.java:61)
        at io.undertow.servlet.handlers.FilterHandler$FilterChainImpl.doFilter(FilterHandler.java:131)
        at org.springframework.web.filter.FormContentFilter.doFilterInternal(FormContentFilter.java:93)
        at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
[...]

PR #195, opened 2020-05-18
Fix NPE happening for ignore filters with skipped ASN or prefix.

lolepezy:
It is to fix 
https://github.com/RIPE-NCC/rpki-validator-3/issues/194


wibisono:
image

I think this change will propagate and NPE might appear on other parts where we combine Slurm and normal ROAs. (For exports or to feed rtr-server)

PR #196, opened 2020-05-20
Fix export-extended

wibisono:
UI still expect export-extended backend to exist, while backend is renamed to roas-only.
I remember we renamed because we don't include SLURM on the extended.

Now this change revert backend API path, and include slurm on extended, without
additional info.

{
    "asn" : "12314",
    "prefix" : "192.168.0.0/24",
    "maxLength" : 24,
    "ta" : "SLURM",
    "notBefore" : "",
    "notAfter" : "",
    "serialNumber" : ""
  }


ties:
Works on my machine. LGTM!

PR #203, opened 2020-05-25
Update changelog

ties:


rpki-validator/Changelog.txt
```diff
@@ -1,4 +1,12 @@
 * Fri May 22 2020 Ties de Kock <tdekock@ripe.net> 3.1
+Security update: Changed permissions for CentOS systemd service files
+
+After a change in our build infrastructure, the CentOS (rpm) artifact contained
+world-writable systemd service files that would allow users with write access
+to the machine to elevate privileges and get local code execution.
+
+Version affected: 3.1-2020.05.08.09.26.49
+
```
wibisono, line 9: Might not be necessary to make this explicit but people have to derive that:
- previous versions were not affected
- it is fixed in 3.1-2020.05.22.11.25

PR #205, opened 2020-05-25
Use Selector to await for channels ready to connect

omuravskiy:


rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
```diff
@@ -134,43 +139,44 @@ private void resolveAsynchronously(String host, int port, Promise<List<InetSocke
...
-                connectionsThread.interrupt();
-
-                try {
-                    connectionsThread.join();
-                } catch (Exception ignore) {
+                    promise.failed(new IOException("Could not connect to " + host));
                 }
-
             } catch (Exception e) {
                 log.warn("Error during looking for happy eyeballs: ", e);
```
lolepezy, line 151: Nice message :)

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
```diff
@@ -273,52 +270,55 @@ private static Runnable dnsLookupRunnable(Name hostname, int queryType, Queue<Op
...
+                    // can't connect - move on
+                    log.info("Error connecting to {}", address);
                 }
-                return true;
+            } else {
+                return false;
             }
-        };
+        }
+        return true;
```
wibisono, line 322: Curious how null could be inserted to the queue, and why we returned true in the open connection from queue in those cases.

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
```diff
@@ -134,43 +139,44 @@ private void resolveAsynchronously(String host, int port, Promise<List<InetSocke
...
+                        closeAnyway(key.channel());
+                    }
+                    closeAnyway(selector);
                 }
             }
         });
     }
 
-    private static void closeAnyway(SocketChannel channel) {
+    private static void closeAnyway(Channel channel) {
```
lolepezy, line 164: This one is not needed since you added the method for Closable.

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
```diff
@@ -273,52 +270,55 @@ private static Runnable dnsLookupRunnable(Name hostname, int queryType, Queue<Op
...
-                            log.debug("Error connecting to {}", address);
-                        } finally {
-                            TimeUnit.MILLISECONDS.sleep(CONNECTION_ATTEMPT_DELAY_MS);
-                        }
-                    } else {
-                        return false;
-                    }
+            final boolean haveMoreAF2 = openConnectionFromQueue(secondAfAddresses, port, selector);
+            connectedSocket = awaitSuccessfulConnection(selector, CONNECTION_ATTEMPT_DELAY_MILLIS);
+            if (connectedSocket != null) return connectedSocket;
```
lolepezy, line 287: Shouldn't this one be the same check for isPresent?

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
```diff
@@ -273,52 +270,55 @@ private static Runnable dnsLookupRunnable(Name hostname, int queryType, Queue<Op
...
+                final InetAddress address = addressOptional.get();
+                try {
+                    log.info("Connecting to {}", address);
+                    final SocketChannel socketChannel = SocketChannel.open();
+                    socketChannel.configureBlocking(false);
+                    socketChannel.register(selector, SelectionKey.OP_CONNECT);
+                    socketChannel.connect(new InetSocketAddress(address, port));
+                } catch (IOException e) {
+                    // can't connect - move on
+                    log.info("Error connecting to {}", address);
```
lolepezy, line 316: log.debug?

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
lolepezy: Well, javadoc literally says " Retrieves and removes the head of this queue, or returns {@code null} if this queue is empty", so it means the queue is empty.

wibisono: For a moment I thought the poll itself is returning Optional (present/empty) but it's the entry type on the queue that was optional. Clear now. Still shouldn't it return false when there is nothing on the queue?

lolepezy: It returns either the element or "nothing", "null" is just java's way to say "nothing" :)

omuravskiy:
> I have run it on mac os and I do see that
> 
> * master branch has thread-runaway behaviour
> * this implementation doesn't have it
> 
> Do you have it running on some linux/bsd anywhere? We can spin up another test instance on AWS as before, I guess.

Yes, I run it on my FreeBSD. It looks fine, although there are lots of HttpClient@... threads that seem to do nothing, just sit there. Not sure if this is a problem.

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
omuravskiy: :)

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
omuravskiy: The empty queue does not mean the end of the queue (or that we went through all addresses).
The signal of the end of the queue is Optional.empty in the queue. This is when we return false (meaning there's nothing more to wait for in this address family). In all other cases we return true.

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
omuravskiy: Good catch! This is a bug.

omuravskiy: yes

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
omuravskiy: Good catch!  This is a bug.

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/HappyEyeballsResolver.java
omuravskiy: Hm, there's whole lot of conditions which would raise an exception on programming error - like connecting on closed socket, on already connected socket, etc. So I think it needs more exposure than DEBUG.

lolepezy:
> Yes, I run it on my FreeBSD. It looks fine, although there are lots of HttpClient@... threads that seem to do nothing, just sit there. Not sure if this is a problem.

It's some jetty pool, I guess, I've seen it multiple times in jvisualvm.

lolepezy:
Looks good to me.

PR #215, opened 2020-06-08
Accept and retrieve trust anchor over https in tal

ties:
Accept trust anchors with https url in tals and use this.

  * Parse and valide the tals that contain https urls
  * Reject other protocols in tals (except file during test)
  * Fetch trust anchor over https as well as rsync

Misc other changes:
  * Made the httpclient into a bean, which is autostarted.
  * Keep rsync timing metrics for both trust anchor certificates and repos


AlexanderBand:
@amreesh confirmed that the TAL for AFRINIC can be fetched at https://rpki.afrinic.net/repository/AfriNIC.cer

For APNIC, currently https://rpki.apnic.net/repository/apnic-rpki-root-iana-origin.cer works but you will want to verify the desired URI with @geeohgeegeeoh before releasing as I understand they intend to have a new canonical location.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/metrics/RsyncMetricsService.java
```diff
@@ -0,0 +1,96 @@
...
+    private MeterRegistry registry;
+
+    private ConcurrentHashMap<Tuple2<String, Integer>, RsyncMetric> rsyncMetrics = new ConcurrentHashMap<>();
+
+    public void update(String uri, int statusDescription, long durationMs) {
+        update(URI.create(uri), statusDescription, durationMs);
+    }
+
+    public void update(URI uri, int statusDescription, long durationMs) {
+        final String relativeURL = uri.resolve("/").toASCIIString();
```
lolepezy, line 66: Did you mean "absoluteURL" or "rootURI" or something like that? I think it's the opposite of "relative". And the same for Http version.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/retrieval/TrustAnchorRetrievalService.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    protected byte[] fetchHttpsTrustAnchorCertificate(URI trustAnchorCertificateURI, ValidationResult validationResult) throws IOException {
+        final long t0 = System.currentTimeMillis();
+        String statusDescription = "unknown";
+
+        try {
+            ContentResponse res = httpClient
+                    .newRequest(trustAnchorCertificateURI)
+                    .header(HttpHeader.USER_AGENT, null)
```
lolepezy, line 105: Why ".header" two times?

rpki-validator/src/main/java/net/ripe/rpki/validator3/config/HttpClientConfig.java
```diff
@@ -0,0 +1,86 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
+
+import javax.annotation.PreDestroy;
+
+@Slf4j
+@Configuration
+public class HttpClientConfig {
```
lolepezy, line 47: It is becoming slightly confusing now.
If this functionality moves away from Http, shall it be renamed to something like HttpStreaming?
And maybe move all http stuff into the same package.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/retrieval/TrustAnchorRetrievalService.java
```diff
@@ -0,0 +1,147 @@
...
+                    .header(HttpHeader.USER_AGENT, String.format("RIPE NCC RPKI Validator/%s", buildInformation.getVersion()))
+                    .send();
+            log.debug("HTTP {} when fetching HTTPS trust anchor from {}", res.getStatus(), trustAnchorCertificateURI);
+            statusDescription = String.valueOf(res.getStatus());
+            if (res.getStatus() != 200) {
+                validationResult.error(ErrorCodes.TRUST_ANCHOR_FETCH, trustAnchorCertificateURI.toASCIIString(), String.format("HTTP %d", res.getStatus()));
```
wibisono, line 111: nit: Instead of formatting status code into "HTTP %d", maybe supply res.getReason().

AlexanderBand:
I received confirmation from APNIC that their root certificate should be fetched via https://tal.apnic.net/apnic.cer. You can update your included APNIC TAL with this URL.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/retrieval/TrustAnchorRetrievalService.java
ties:
```java
    /**
     * <p>Adds the given {@code value} to the specified {@code header}.</p>
     * <p>Multiple calls with the same parameters will add multiple values;
     * use the value {@code null} to remove the header completely.</p>
     * ...
     */
```

ties: Turns out we have been setting the user agent in a strange fashion all the time, I'll refactor it to use a different method.

rpki-validator/src/main/java/net/ripe/rpki/validator3/config/HttpClientConfig.java
ties: Good suggestion!

ties:
> I received confirmation from APNIC that their root certificate should be fetched via https://tal.apnic.net/apnic.cer. You can update your included APNIC TAL with this URL.

Great! I'll update this

AlexanderBand:
> > I received confirmation from APNIC that their root certificate should be fetched via https://tal.apnic.net/apnic.cer. You can update your included APNIC TAL with this URL.
> 
> Great! I'll update this

We just got word that the ARIN TAL is now also available with an HTTPS URI. I'm not sure what procedure you have for re-fetching/updating the TAL, but the new one is published on their website: https://www.arin.net/resources/manage/rpki/arin.tal

geeohgeegeeoh:
We prefer the URL https://tal.apnic.net/apnic.tal be used for TAL but note
it refers to the canonical publication path of the cert. We include
https://tal.apnic.net/apnic.cer as a path, but this is not the canonical
path to the actual trust anchor certificate.

I suggest at this time only using the TAL filepath URL.

PR #219, opened 2020-06-12
Fallback to rsync from https when trust anchor certificate can not be loaded

ties:
Quick and slightly dirty implementation of fallback behaviour. Would like it if the implementation is slightly cleaner.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/TrustAnchorValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/TrustAnchorValidationService.java
@@ -29,6 +29,7 @@
  */
 package net.ripe.rpki.validator3.domain.validation;
 
+import com.google.common.base.Joiner;
 import com.google.common.collect.Sets;
 import com.google.common.io.Files;
 import lombok.extern.slf4j.Slf4j;
@@ -66,6 +67,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.security.GeneralSecurityException;
 import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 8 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/data/TrustAnchor.java
```diff
@@ -98,4 +101,11 @@ public X509ResourceCertificate getCertificate() {
     public void markInitialCertificateTreeValidationRunCompleted() {
         this.initialCertificateTreeValidationRunCompleted = true;
     }
+
+    public List<URI> getLocationsByPreference() {
+        return locations.stream()
+            .sorted(Comparator.naturalOrder())
```
wibisono, line 107: Just to confirm the intent, sorting in the ends put https in front of rysnc, right?

ties:
Yup, dat is the intent.

Perhaps I should document that instead of it being implicit.

rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/data/TrustAnchor.java
ties: Sorting would re-order urls with the same scheme alphabetically. I changed this.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/TrustAnchorValidationService.java
```diff
@@ -121,41 +123,41 @@ public void validate(long trustAnchorId) {
             return new TrustAnchorValidationRun(trustAnchorRef, trustAnchor.getLocations().get(0));
         });
 
+        final ValidationLocation trustAnchorValidationLocation = new ValidationLocation(validationRun.getTrustAnchorCertificateURI());
+        ValidationResult validationResult = ValidationResult.withLocation(trustAnchorValidationLocation);
+
         boolean updatedTrustAnchor = false;
+        final List<URI> rankedTrustAnchorLocations = trustAnchor.getLocationsByPreference();
         try {
+            boolean loadedAtLeastOneTACertificate = false;
```
lolepezy, line 132: Can you hide this logic into a separate method? Here we basically only want to know if TA cert was fetched or not, without all the details.

PR #220, opened 2020-06-12
Use S6 container overlay so container exits properly

ties:
* Use S6 overlay for process management
  * Use ADD instead of build so you can use URLs in build
  * Use RPKI user
  * Chown data and config dir on startup
  * Clean up xodus lockfile on startup


ties:
Just for documentation: s6 overlay v2.0.0.1 caused an error when starting.

`standard_init_linux.go:211: exec user process caused "no such file or directory"`

PR #222, opened 2020-06-16
Move application version to different property file

### LINKED ISSUE #216: Application version is in config file that is adjusted by the user
As a user I want my settings to be retained after upgrading the application. However, right now the application version is in `/etc/rpki-validator-3/application.properties` for the Debian package. This causes a popup about this file on every upgrade. The version should be in the file with the defaults.

`/etc/rpki-validator-3/application.properties` is marked as a conffile -> it is user adjustable.

https://www.debian.org/doc/debian-policy/ap-pkg-conffiles.html

ties:
Fixes #216

### DIFF
--- a/rpki-rtr-server/src/main/resources/packaging/generic/conf/application-defaults.properties
+++ b/rpki-rtr-server/src/main/resources/packaging/generic/conf/application-defaults.properties
@@ -42,6 +42,9 @@ server.port=8081
 #
 rtr.server.address=localhost
 
+# Used in the user agent when fetching RRDP
+validator.version=@build.version@-@build.release@
+
 #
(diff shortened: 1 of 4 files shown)


ties:
Note that there is just one `application-defaults.properties` that is used by all the builds:
```
~/src/ripe/rpki/rpki-validator-3 fix/application-version-in-non-config-file
❯ find . -name application-default\*
./rpki-rtr-server/target/classes/packaging/generic/conf/application-defaults.properties
./rpki-rtr-server/src/main/resources/packaging/generic/conf/application-defaults.properties
./rpki-validator/target/generic-conf/application-defaults.properties
./rpki-validator/src/main/resources/packaging/generic/conf/application-defaults.properties
```

PR #224, opened 2020-06-17
Log happy eyeballs resolver at debug level

ties:
Makes it a _lot_ less noisy!


ties:
@omuravskiy this is just for debugging, isn't it? It is the majority of the output for the docker container so I would like to strip it.

omuravskiy:
> @omuravskiy this is just for debugging, isn't it? It is the majority of the output for the docker container so I would like to strip it.

Yeah, it's fine. Although I would keep second one at INFO, to annoy people about their broken IPv6.

ties:
IPv6 is almost always broken in docker though.

Perhaps we can log once per ip?

Issue #162, opened 2020-03-30
An expired or missing CRL doesn't result in the manifest being considered invalid


job:
This issue was also informally reported to RIPE NCC staff in late February 2020.

It appears that RIPE NCC validator doesn't consider an expired CRL cause to consider the associated manifest and underlaying objects invalid. It logs a warning, but does not consider this situation erroneous.

Theoretically a replay attack can be executed by making the CRL unavailable in a monkey-in-the-middle scenario, or presenting the validator with old data. The RPKI data download procedure in RIPE NCC's Validator must assume it is collecting untrusted data, which must be verified using the certificate verification model where an expired CRL means that none of the X509 objects pointing to the CRL can be validated or considered valid.

Some more (hard to read) context is available in the RPKI standards forum at IETF via https://mailarchive.ietf.org/arch/msg/sidrops/BKbYujWcr4qWJOahFoBqgZHrFX8/ and specifically https://mailarchive.ietf.org/arch/msg/sidrops/6bHseJeNPSsK0lebbNTDGFKVYsA/

If the CRL is expired, the manifest & directory should not be considered eligible for further processing and the associated VRPs should not be emitted to the RTR process. Manifests and CRLs compliment each other, manifests are not a replacement for CRLs.

The manifest specification assumes a sensible policy regarding CRL expiration is applied. Since RPKI does not have any other certificate revocation mechanisms other than CRLs (there is no OSCP and other tricks in the specs) - I believe that ignoring the CRLs expiration date opens up RIPE NCC's validator for replay attacks which can have adverse operational consequences (aka an attacker has the ability to flip BGP routes from `valid` to `invalid`)

To be transparent: I'm filing this bug publicly so I can reference this later on in a write-up about how these validation issues were handled.

lolepezy:
It's been addressed in the latest release
https://github.com/RIPE-NCC/rpki-validator-3/releases/tag/3.1-2020.07.06.14.28
it would still require explicitly setting the property `rpki.validator.strict-validation = true` but the functionality is there.

Issue #158, opened 2020-03-23
Missing files don't result in a manifest being considered invalid


job:
In a MITM scenario where an attacker intercepts and manipulates the rsync channel (for example strategically withholding certain .roa files from the view of the validator being attacked), the resulting set of VRPs will be incomplete and can cause severe operational issues.

When a manifest is valid (manifest is parsable, CRL exists is valid (also not expired), and manifest is signed with keys not revoked by the CRL), *and* references files which do not exist in the repository at hand, the publication point should be considered compromised.

So in the case of APNIC where an End User (self-hosted) RPKI publication point misses a few .roa files, the validator can proceed to consider all data from all RPs it could reach eligeble for further validation, except any data from the publication point where files were missing.

In other words: if one or a few files are missing from the repository, the repository should be considered 'down', and no attempt should be made to start guessing what can be salvaged and what not.

job:
To offer an example:

In manifest `1-tcQDnftkRnWbiMhu2cR1-dgmCs.mft`, a number of ROAs and a CRL are referenced:

```
1-tcQDnftkRnWbiMhu2cR1-dgmCs.crl
IM3xMiMU1QChuWMOGgnbDwYolrU.roa
LkKeUPYrfgzjsOIejLjsHGk44cU.roa
LkKeUPYrfgzjsOIejLjsHGk44cU.roa
fIeCcC8KpdJQd-olU2APdxNZkyA.roa
r5vtD4isKhTzGXNGulSnZ7XCS04.roa
r7TSyWn_GbYPjNWvt4r5ewSNAsk.roa
```

If an attacker withholds all but `LkKeUPYrfgzjsOIejLjsHGk44cU.roa`, so the MITM does the equivalent of:

```
cd repository/DEFAULT/3e/01d411-d915-4277-8fe2-76b0dda2bf3e/1/
rm LkKeUPYrfgzjsOIejLjsHGk44cU.roa \
    fIeCcC8KpdJQd-olU2APdxNZkyA.roa \
    IM3xMiMU1QChuWMOGgnbDwYolrU.roa \
    NpI_Uj3OZb6jvwfUTX0-eOk9QV4.roa
```

and the validator does *not* consider `1-tcQDnftkRnWbiMhu2cR1-dgmCs.mft` in its entirety invalid due to one or more missing `.roa` or `.crl` files, in this specific example the remaining ROA will render all BGP announcements equal to or covered under `80.128.0.0/11` `invalid`.

[...]

PR #226, opened 2020-06-18
Expose strict properties

wibisono:
Related to #162 #158

### DIFF
--- a/rpki-validator/src/main/resources/application.properties
+++ b/rpki-validator/src/main/resources/application.properties
@@ -92,7 +92,7 @@ management.endpoints.enabled-by-default=false
 management.endpoint.prometheus.enabled=true
 management.endpoints.web.exposure.include=prometheus
 
-# Enabling strict validation will not allow stale manifest, stale crl, and reject manifest with entries referring to non-existent objects.
+# Enabling strict validation will reject stale manifest, stale crls, and manifest with entries referring to non-existent objects.
 rpki.validator.strict-validation=false
 
 # This disables RRDP, so the validator only uses rsync to download repositories.
--- a/rpki-validator/src/main/resources/application.properties
+++ b/rpki-validator/src/main/resources/application.properties
@@ -92,3 +92,5 @@ management.endpoints.enabled-by-default=false
 management.endpoint.prometheus.enabled=true
 management.endpoints.web.exposure.include=prometheus
 
+# Enabling strict validation will not allow stale manifest, crl, and will complain about non existing manifest entry.
+rpki.validator.strict-validation=false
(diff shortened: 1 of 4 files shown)


rpki-validator/src/main/resources/application.properties
```diff
@@ -92,3 +92,8 @@ management.endpoints.enabled-by-default=false
 management.endpoint.prometheus.enabled=true
 management.endpoints.web.exposure.include=prometheus
 
+# Enabling strict validation will not allow stale manifest, stale crl, and reject manifest with non existing entry.
```
lolepezy, line 95: "with entries referring to non-existent objects"?

rpki-validator/src/main/resources/application.properties
```diff
@@ -92,3 +92,8 @@ management.endpoints.enabled-by-default=false
...
+rpki.validator.strict-validation=false
+
+# Enabling rsync only will ignore RRDP repositories and only use rsync
```
lolepezy, line 98: Slightly too wordy. 
"This disables RRDP, so the validator only uses rsync to download repositories."?

rpki-validator/src/main/resources/application.properties
```diff
@@ -92,3 +92,8 @@ management.endpoints.enabled-by-default=false
 management.endpoint.prometheus.enabled=true
 management.endpoints.web.exposure.include=prometheus
 
+# Enabling strict validation will not allow stale manifest, stale crl, and reject manifest with entries referring to non-existent objects.
+rpki.validator.strict-validation=false
+
+# This disables RRDP, so the validator only uses rsync to download repositories.
+rpki.validator.rsync-only=false
```
ties, line 99: Just checking: there is no corresponding rrdp only?

rpki-validator/src/main/resources/application.properties
```diff
@@ -92,3 +92,8 @@ management.endpoints.enabled-by-default=false
 management.endpoint.prometheus.enabled=true
 management.endpoints.web.exposure.include=prometheus
 
+# Enabling strict validation will not allow stale manifest, stale crl, and reject manifest with entries referring to non-existent objects.
```
ties, line 95: Do stale manifests and crls also get rejected?

wibisono: Yes

rpki-validator/src/main/resources/application.properties
wibisono: Nope.

lolepezy: > Just checking: there is no corresponding rrdp only?

We probably don't need "rrdp only" because choosing rrdp is the default behaviour.
And we do it for the testbed, so it should be enough to have a "default" and "rsync only" versions.

ties: It does not feel super useful, but it would make it possible to check what goes missing without fallback.

Routinator covers this way of checking.

rpki-validator/src/main/resources/application.properties
ties, line 95:
```suggestion
# Enabling strict validation will reject stale manifest, stale crls, and manifest with entries referring to non-existent objects.
```

PR #230, opened 2020-07-03
Don't store stack trace for transaction

erikrozendaal:
Gathering the stack trace is very expensive. Remove this speeds up the validation process by about 10%-30%.


lolepezy:
It would also make sense to remove the storing structures for these stack traces and the "api/xodus/tx" end-point.

PR #231, opened 2020-07-06
Fix/npe when serialising null param

wibisono:
Fixing this failing test:
https://github.com/ties/rpki-validator-3/commit/5cb3375ced110d52bee30857630cb4396379cbe0


rpki-validator/src/test/java/net/ripe/rpki/validator3/storage/encoding/custom/CodersTest.java
```diff
@@ -69,5 +69,12 @@ public void testBigInteger(BigInteger b) {
     public void testListString(List<String> b) {
         assertEquals(b, Coders.fromBytes(Coders.toBytes(b, Coders::toBytes), Coders::toString));
     }
-
+    @Property
```
ties, line 72:
```suggestion

    @Property
```

PR #233, opened 2020-07-08
Use Xodus logcache based on soft references

ties:
Makes xodus use soft references for it's cache instead of a fixed
partition. This is documented in [0]. It is said to solve an issue [1]
where sequential writes fail due to out of memory.

[0]: https://youtrack.jetbrains.com/issue/XD-799
[1]: https://youtrack.jetbrains.com/issue/XD-768


ties:
After we get a beta out I'll track the change in behaviour (both memory usage and GC time) of my instance.

PR #234, opened 2020-07-14
Reduce BGP preview memory usage

erikrozendaal:
Shares ROA prefix instances between `BgpPreviewService` and `ValidatedRpkiObjects` and stores the `BgpPreviewEntry` data more efficiently. Together heap usage is reduced by about 75 MB.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpPreviewService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpPreviewService.java
@@ -112,8 +112,6 @@ public class BgpPreviewService {
 
     private static IpRange DEFAULT_IPV4_ROUTE = IpRange.parse("0.0.0.0/0");
     private static IpRange DEFAULT_IPV6_ROUTE = IpRange.parse("::/0");
-    private static BigInteger TWO_64 = BigInteger.ONE.shiftLeft(64);
-    private static BigInteger TWO_64_MINUS_1 = TWO_64.subtract(BigInteger.ONE);
 
     @lombok.Value(staticConstructor = "of")
     public static class BgpPreviewResult {
@@ -159,36 +157,24 @@ public class BgpPreviewService {
         List<ValidatingRoa> filteredRoas;
     }
 
-    @lombok.Value
-    public static class BgpPreviewEntry {
+    @lombok.Getter
+    @lombok.EqualsAndHashCode
+    @lombok.AllArgsConstructor
+    @lombok.ToString
+    public static abstract class BgpPreviewEntry {
         // store it in memory optimised way, as we are going to store a lot of these objects in memory
-        int origin;
-        short prefixLength; // negative for IPv6
-        long prefixHi;
-        long prefixLo;
-        Validity validity;
+        protected final int origin;
+        protected final Validity validity;
 
         public static BgpPreviewEntry of(Asn origin, IpRange prefix, Validity validity) {
-            short prefixLength;
-            long prefixHi;
-            long prefixLo;
             switch (prefix.getType()) {
                 case IPv4:
-                    prefixLength = (short) prefix.getPrefixLength();
-                    prefixHi = 0;
-                    prefixLo = ((Ipv4Address) prefix.getStart()).longValue();
-                    break;
-                case IPv6: {
-                    BigInteger value = prefix.getStart().getValue();
-                    prefixLength = (short) (-prefix.getPrefixLength() - 1);
-                    prefixHi = value.shiftRight(64).and(TWO_64_MINUS_1).longValue();
-                    prefixLo = value.and(TWO_64_MINUS_1).longValue();
-                    break;
-                }
+                    return BgpPreviewEntry4.of(origin, validity, prefix);
+                case IPv6:
+                    return BgpPreviewEntry6.of(origin, validity, prefix);
                 default:
                     throw new IllegalArgumentException("invalid IP prefix type: " + prefix.getType());
             }
-            return new BgpPreviewEntry((int) origin.longValue(), prefixLength, prefixHi, prefixLo, validity);
         }
 
         private static Predicate<BgpPreviewEntry> matches(SearchTerm searchTerm) {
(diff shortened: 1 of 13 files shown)


erikrozendaal:
The last three commits reduce memory usage by another 25 MB or so. The in-memory data we keep (validated ROA prefix, BGP entries, etc) are now less than 100 MB total.

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpPreviewService.java
```diff
@@ -238,16 +229,82 @@ public static BgpPreviewEntry of(Asn origin, IpRange prefix, Validity validity)
...
+
+    @lombok.Value
+    @lombok.EqualsAndHashCode(callSuper = true)
+    @lombok.ToString(callSuper = true)
+    public static class BgpPreviewEntry6 extends BgpPreviewEntry {
+        private static BigInteger TWO_64 = BigInteger.ONE.shiftLeft(64);
+        private static BigInteger TWO_64_MINUS_1 = TWO_64.subtract(BigInteger.ONE);
+
+        short prefixLength;
+        long prefixHi;
```
ties, line 276: You're splitting the 128 bits of the prefix into two longs. It would be clearer for me if there was a small piece of documentation here that indicated that is what this class is doing :)

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpPreviewService.java
```diff
@@ -238,16 +229,82 @@ public static BgpPreviewEntry of(Asn origin, IpRange prefix, Validity validity)
...
+            this.prefixLength = prefixLength;
+            this.prefixHi = prefixHi;
+            this.prefixLo = prefixLo;
+        }
+
+        public static BgpPreviewEntry6 of(Asn origin, Validity validity, IpRange prefix) {
+            BigInteger value = prefix.getStart().getValue();
+            short prefixLength = (short) prefix.getPrefixLength();
+            long prefixHi = value.shiftRight(64).and(TWO_64_MINUS_1).longValue();
+            long prefixLo = value.and(TWO_64_MINUS_1).longValue();
```
ties, line 290: BigInteger.longValue: "Converts this BigInteger to a long. This conversion is analogous to a narrowing primitive conversion from long to int as defined in section 5.1.3 of The Java™ Language Specification: if this BigInteger is too big to fit in a long, only the low-order 64 bits are returned. Note that this conversion can lose information about the overall magnitude of the BigInteger value as well as return a result with the opposite sign."

Is the masking needed?

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpPreviewService.java
```diff
@@ -238,16 +229,82 @@ public static BgpPreviewEntry of(Asn origin, IpRange prefix, Validity validity)
...
 
-        public Asn getOrigin() {
-            return new Asn(origin);
+    }
+
+    @lombok.Value
+    @lombok.EqualsAndHashCode(callSuper = true)
+    @lombok.ToString(callSuper = true)
+    public static class BgpPreviewEntry6 extends BgpPreviewEntry {
+        private static BigInteger TWO_64 = BigInteger.ONE.shiftLeft(64);
```
ties, line 272: A 1 + 64 zeroes

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpPreviewService.java
```diff
@@ -238,16 +229,82 @@ public static BgpPreviewEntry of(Asn origin, IpRange prefix, Validity validity)
...
+        private static BigInteger TWO_64_MINUS_1 = TWO_64.subtract(BigInteger.ONE);
```
ties, line 273: 64 1's

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpPreviewService.java
```diff
@@ -238,16 +229,82 @@ public static BgpPreviewEntry of(Asn origin, IpRange prefix, Validity validity)
...
         }
 
         public IpRange getPrefix() {
-            return prefix.toIpRange();
+            BigInteger prefix = BigInteger.valueOf(prefixHi);
+            if (prefixHi < 0) {
+                prefix = prefix.add(TWO_64);
+            }
+            prefix = prefix.shiftLeft(64).add(BigInteger.valueOf(prefixLo));
+            if (prefixLo < 0) {
```
ties, line 300: My two complement's rusty: Would be easier for me to read when you perform the `add(TWO_64) first (if needed) and then add if to the shifted prefix.

As a bonus this may be negibly faster since `prefixLo` is shorter than prefix post-shifting.

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpPreviewService.java
```diff
@@ -184,24 +189,33 @@ public int getEffectiveLength() {
...
+            long prefixHi;
+            long prefixLo;
+            switch (prefix.getType()) {
+                case IPv4:
+                    prefixLength = (short) prefix.getPrefixLength();
+                    prefixHi = 0;
+                    prefixLo = ((Ipv4Address) prefix.getStart()).longValue();
+                    break;
+                case IPv6: {
+                    BigInteger value = prefix.getStart().getValue();
```
wibisono: No idea which approach is better, but here's another ipv6 stored as long implementation, and function that initialize it from BigInteger. Not quite sure why it goes first to byte buffer:
https://github.com/sargun/java-ipv6/blob/master/src/main/java/com/googlecode/ipv6/IPv6Address.java#L157

erikrozendaal: I prefer shifting + conversion, since it more closely reflects the required math. Performance is probably very similar and not critical anyway.

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpPreviewService.java
erikrozendaal: Removed the implicit masking. I was hoping to use `longValueExact`, but that probably doesn't work when the 64-bit unsigned value doesn't fit in a 64-bit signed value (resulting in a negative value).

rpki-validator/src/main/java/net/ripe/rpki/validator3/api/bgp/BgpPreviewService.java
erikrozendaal: Added documentation.

PR #235, opened 2020-07-17
Reduce Xodus memory usage by using a smaller log cache page size

erikrozendaal:
The default log cache page size in Xodus is 64 KB. This is useful when
data that is used together is stored near eachother. However, in RPKI
most objects are addressed by the SHA-256 value of the object. This is
basically "random" so we do not get any advantage when loading objects
that are nearby to the object we're currently interested in.

Reducing this log page cache size to 2 KB ensures almost all RPKI
objects still fit and we do not waste CPU and memory loading
co-located objects that are very unlikely to be needed.

With this setting I've been able to run the RPKI validator with the
five RIR trust anchors for more than 13 hours using 512 MB maximum
heap size without issue. Much better than the current required 1536 MB
max heap size!

JVM options used: `-verbose:gc -XX:+ExitOnOutOfMemoryError -XX:+HeapDumpOnOutOfMemoryError -Xms512M -Xmx512M`

CPU and heap usage:

image

Heap dump overview:

image


ties:
Great find!

PR #236, opened 2020-07-20
Add explanation.

lolepezy:


ui-validator/src/app/announcement-preview/announcement-preview.component.html
```diff
@@ -81,6 +81,9 @@ <h4 class='modal-title'>{{'AnnouncementPreview.TITLE_MODAL_FORM' | translate}}</
   <form #asnPrefixForm='ngForm'
         (ngSubmit)='onSubmitModal(asnPrefixForm.value)'>
     <div class='modal-body'>
+      <p>This page shows the impact of the ROAs available at the moment at the provided (ASN+Prefix) pair.
```
ties, line 84: I don't think it's "at", but "for" would be more suitable? 

And where are the ROAs available? It refers to the current state of the validator? If so we can be more explicit (sorry, do not have a direct suggestion here).

lolepezy: Yes, indeed. Adjusted the text a little.

PR #237, opened 2020-07-20
Accumulate ROA prefixes and router certificates during TA validation

erikrozendaal:
This saves a round-trip to the Xodus store to retrieve ROAs and router certificates after a successful validation run.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/ValidatedRpkiObjects.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/ValidatedRpkiObjects.java
@@ -34,9 +34,11 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
 import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
+import net.ripe.rpki.commons.crypto.CertificateRepositoryObject;
+import net.ripe.rpki.commons.crypto.cms.roa.RoaCms;
+import net.ripe.rpki.commons.crypto.cms.roa.RoaPrefix;
 import net.ripe.rpki.commons.crypto.x509cert.X509CertificateUtil;
 import net.ripe.rpki.commons.crypto.x509cert.X509RouterCertificate;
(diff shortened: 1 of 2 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/ValidatedRpkiObjects.java
```diff
@@ -291,4 +258,70 @@ private void notifyListeners() {
         Locks.locked(dataLock.readLock(), () ->
             listeners.forEach(listener -> listener.accept(validatedObjectsByTrustAnchor.values())));
     }
+
+    public static class Accumulator {
+        private final TrustAnchorData trustAnchorData;
+        private final List<Key> validatedObjectKeys = Collections.synchronizedList(new ArrayList<>());
```
ties, line 264: I get a bit skittish when i see Collections.synchronizedList`. Since it seems like we only need `size` for logging (and could return it from `markReachable`): Can we use a wait-free datastructure/`ConcurrentLinkedQueue` here?

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/ValidatedRpkiObjects.java
```diff
@@ -291,4 +258,70 @@ private void notifyListeners() {
...
+                        pkInfo
+                ));
+            }
+        }
+
+        public boolean isEmpty() {
+            return validatedObjectKeys.isEmpty();
+        }
+
+        public int size() {
```
ties, line 307: Only works because it's O(1) on `ArrayList`, would be bad with my proposed change.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/ValidatedRpkiObjects.java
erikrozendaal: Synchronized list was used here because it is the data structure currently used and we use `parallelStream` in the validation process. A cleaner approach would be to stream the results back and use `flatMap` to collect all data from the various threads. Not sure if it matters much for performance though.

PR #245, opened 2020-07-29
Only store value if it is different from old value

erikrozendaal:
We were already loading the old value from the store anyway so comparing it with the new value should be cheap. Hopefully this reduces the growth of the (log-structured) Xodus data store.


erikrozendaal:
This mainly occurs for rsync objects, as these are always re-inserted into the data store.

lolepezy:
Shouldn't we introduce some "exists" method in IxMap that would be very cheap and not extract anything at all? So it could be used in both rsync and rrdp services?

PR #246, opened 2020-07-29
Rsync batched loading

erikrozendaal:
Similar to the RRDP snapshot streaming parsing, this loads objects in batches of 1 MB. Two passes are made over the file system directory, first for all non-manifest objects and the second for all the manifests. This to ensure that every object referenced from a manifest will be in the data store when validating.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
@@ -78,9 +78,6 @@ import java.util.HashSet;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ExecutorCompletionService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Stream;
 
 @Service
@@ -349,10 +346,6 @@ public class RpkiRepositoryValidationService {
                                     ValidationResult validationResult,
                                     Set<String> storedObjectKeys,
                                     RpkiRepository repository) throws IOException {
-
-        final AtomicInteger workCounter = new AtomicInteger(0);
-        final int threshold = 10;
-
         Files.walkFileTree(targetDirectory.toPath(), new SimpleFileVisitor<Path>() {
             private URI currentLocation = URI.create(repository.getLocationUri());
 
@@ -404,12 +397,8 @@ public class RpkiRepositoryValidationService {
                     boolean exists = storedObjectKeys.contains(key);
 
                     if (!exists) {
-                        if (workCounter.get() > threshold) {
-                            storeObject(tx, validationRun, storedObjectKeys);
-                            workCounter.decrementAndGet();
-                        }
-                        asyncCreateObjects.submit(() -> RpkiObjectUtils.createRpkiObject(location, content));
-                        workCounter.incrementAndGet();
+                        Either<ValidationResult, Pair<String, RpkiObject>> maybeRpkiObject = RpkiObjectUtils.createRpkiObject(location, content);
+                        storeObject(tx, validationRun, storedObjectKeys, maybeRpkiObject);
                     } else {
                         rpkiObjects.addLocation(tx, Key.of(key), location);
                     }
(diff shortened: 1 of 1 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
```diff
@@ -390,69 +394,67 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
...
-                        if (workCounter.get() > threshold) {
-                            storeObject(tx, validationRun, existingObjectsKeys);
-                            workCounter.decrementAndGet();
-                        }
-                        asyncCreateObjects.submit(() -> RpkiObjectUtils.createRpkiObject(location, content));
-                        workCounter.incrementAndGet();
-                    } else {
-                        rpkiObjects.addLocation(tx, Key.of(key), location);
+                    final byte[] content = Files.readAllBytes(file);
+                    pendingObjects.add(Pair.of(location, content));
```
ties, line 409: Since `<uri, blob>` is kind of common in the data model of the validator, it may be worth it to use a value class?

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
```diff
@@ -333,26 +339,24 @@ private RpkiRepository findDownloadedParentRepository(Map<URI, RpkiRepository> f
...
-                                    Set<String> existingObjectsKeys,
-                                    RpkiRepository repository) throws IOException {
-
-        final AtomicInteger workCounter = new AtomicInteger(0);
-        final int threshold = 10;
+                                    Set<String> storedObjectKeys,
+                                    RpkiRepository repository,
+                                    Predicate<RepositoryObjectType> typePredicate) throws IOException {
+        AtomicInteger pendingObjectsBytes = new AtomicInteger(0);
+        List<Pair<String, byte[]>> pendingObjects = new ArrayList<>(1000);
```
lolepezy, line 359: Would it make sense to put filenames here instead of the objects and read them later?

lolepezy: Also, since `asyncCreateObjects` is gone, how slower it is now?
I seems we save a lot of heap, but make it slower, how significant is the speed difference?

lolepezy:
Also, I'm thinking is all the effort to read non-manifests first and then manifests actually worth the hassle. Can we guarantee that tree validation never runs for a TA if one of its repositories is being updated? Some sort of lock would do.

You could drop all the double-traversal code from both rsync and rrdp services.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
erikrozendaal: Some timings from the log output. These timings are unreliable, so should only be used as a rough indication:

Current code on master:

```
Stored 11654 objects [...] in 2723ms
Stored 8381 objects [...] in 1503ms
```

Two FS traversal phases, no executor pool:

```
Stored 11651 objects [...] in 6797ms
Stored 8381 objects [...] in 4023ms
```

Singe FS traversal was not measurably different.

Two FS traversal phases with parallel stream for RPKI object conversion:

```
Stored 11651 objects [...] in 2964ms
Stored 8381 objects [...] in 1653ms
```

Above but first check for existence using SHA-256 hash without parsing the object at all:

```
Stored 11651 objects [...] in 2168ms
Stored 8381 objects [...] in 1088ms
```

For now I'll commit the parallel stream version as that is probably fast enough without adding additional code for existence checking.

erikrozendaal:
> Also, I'm thinking if all the effort to read non-manifests first and then manifests actually is worth the hassle. Can we guarantee that tree validation never runs for a TA if one of its repositories is being updated? Some sort of lock would do.

I think this is harder to get right. Also, unless we perform the locking in the database a restart of the validator will allow validation to run on a partly imported repository. Finally, the actual `rsync` command should be able to run in parallel to the validation job. Only importing the objects from the FS into the database should then have locking.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
lolepezy: Cool!

PR #254, opened 2020-08-06
Corrupted snapshot must be monitored and used to trigger alert

wibisono:
Some of these thrown RrdpException needs to be monitored. Especially corrupt rrdp snapshot.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/RrdpServiceImpl.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/RrdpServiceImpl.java
@@ -225,7 +225,7 @@ public class RrdpServiceImpl implements RrdpService {
                     in,
                     (snapshotInfo) -> {
                         if (!notification.sessionId.equals(snapshotInfo.getSessionId())) {
-                            rrdpMetrics.update(notification.snapshotUri, ErrorCodes.RRDP_WRONG_SNAPSHOT_HASH);
+                            rrdpMetrics.update(notification.snapshotUri, ErrorCodes.RRDP_WRONG_SNAPSHOT_SESSION);
                             throw new RrdpException(ErrorCodes.RRDP_WRONG_SNAPSHOT_SESSION, "Session id of the snapshot (" + snapshotInfo.getSessionId() +
                                     ") is not the same as in the notification file: " + notification.sessionId);
                         }
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/RrdpServiceImpl.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/RrdpServiceImpl.java
@@ -225,6 +225,7 @@ public class RrdpServiceImpl implements RrdpService {
                     in,
                     (snapshotInfo) -> {
                         if (!notification.sessionId.equals(snapshotInfo.getSessionId())) {
+                            rrdpMetrics.update(notification.snapshotUri, ErrorCodes.RRDP_WRONG_SNAPSHOT_HASH);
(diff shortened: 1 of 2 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/RrdpServiceImpl.java
```diff
@@ -261,6 +263,7 @@ private Delta readDelta(Notification notification, DeltaInfo di) {
         final byte[] deltaBody = rrdpClient.getBody(di.getUri());
         final byte[] deltaHash = Sha256.hash(deltaBody);
         if (!Arrays.equals(Hex.parse(di.getHash()), deltaHash)) {
+            rrdpMetrics.update(notification.snapshotUri, ErrorCodes.RRDP_WRONG_DELTA_HASH);
             throw new RrdpException(ErrorCodes.RRDP_WRONG_DELTA_HASH, "Hash of the delta file " + di + " is " + Hex.format(deltaHash) +
```
ties, line 267: These ones aren't caught at a higher level?

wibisono: Actually caught but reported as nothing alarming (fallback to snapshot), while this is seems to be something I'd like to monitor (corrupt hash).  Basically what we are reporting before this PR are metrics falling back to snapshot or new session. I need something to raise the alarm.

It is still not this one, but the IO Exception that will trigger alert.

rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/RrdpServiceImpl.java
```diff
@@ -225,6 +225,7 @@ private int processDownloadedSnapshot(RpkiRepository rpkiRepository, RpkiReposit
                     in,
                     (snapshotInfo) -> {
                         if (!notification.sessionId.equals(snapshotInfo.getSessionId())) {
+                            rrdpMetrics.update(notification.snapshotUri, ErrorCodes.RRDP_WRONG_SNAPSHOT_HASH);
```
erikrozendaal, line 228: Shouldn't this be `ErrorCodes.RRDP_WRONG_SNAPSHOT_SESSION`? Or place this at line 185 where the hash comparison failed?

wibisono: It should be RRDP_WRONG_SNAPSHOT_SESSION, updated.

rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/RrdpServiceImpl.java
```diff
@@ -253,6 +254,7 @@ private int processDownloadedSnapshot(RpkiRepository rpkiRepository, RpkiReposit
 
             return counter.get();
         } catch (IOException e) {
+            rrdpMetrics.update(notification.snapshotUri, ErrorCodes.RRDP_CORRUPTED_SNAPSHOT);
             throw new RrdpException("Couldn't read snapshot: ", e);
```
lolepezy, line 258: Why don't we throw exception with the appropriate error code and update the metrics where we catch the exception? I think it would save some repetitive code.

lolepezy (approved):

rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/RrdpServiceImpl.java
wibisono: In some cases metrics were updated while we're not throwing error (L175, L168), so I'll keep it this way.

PR #257, opened 2020-08-19
Http streaming timeouts and sequental RRDP downloading

erikrozendaal:
See if this fixes a problem with "hanging" RPKI validators.

Normal case there will be very few deltas for a single RRDP
repository. When there are many it is useful to ensure that the first
delta is downloaded and processed first as it doesn't make sense to
download additional ones when the first one is stick for some reason.


ties:
LGTM!

PR #258, opened 2020-08-20
Streaming RRDP delta parser

erikrozendaal:
Use temporary file and streaming to process RRDP deltas to avoid reduce peak memory usage. 

No longer limit the number of concurrent RRDP deltas using a separate thread pool, since memory usage should not be a concern anymore.


rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/RrdpServiceImpl.java
```diff
@@ -259,30 +248,81 @@ private int processDownloadedSnapshot(RpkiRepository rpkiRepository, RpkiReposit
...
+    }
+
+    private int processDownloadedDelta(RpkiRepository rpkiRepository, RpkiRepositoryValidationRun validationRun, Notification notification, DeltaInfo di, Path deltaPath, Predicate<RepositoryObjectType> typePredicate) {
+        try (InputStream in = new FileInputStream(deltaPath.toFile())) {
+            AtomicInteger counter = new AtomicInteger(0);
+            AtomicInteger pendingObjectsBytes = new AtomicInteger(0);
+            List<DeltaElement> pendingObjects = new ArrayList<>(1000);
+            Runnable commitPendingObjects = () -> {
+                storage.readTx0(tx -> verifyDeltaIsApplicable(tx, pendingObjects));
+                storage.writeTx0(tx -> counter.addAndGet(
```
lolepezy, line 287: This one, we already decided, should be adjusted in such a way that parallelStream doesn happen inside of writeTx

lolepezy (changes requested):

rpki-validator/src/main/java/net/ripe/rpki/validator3/rrdp/RrdpServiceImpl.java
erikrozendaal: Did that in a separate pull request https://github.com/RIPE-NCC/rpki-validator-3/pull/259

lolepezy (approved):

wibisono (approved):
Seems sensible,  the other PR with store lock seems to me more closer to real solution. So I might test the other one first and check maybe it is the real solution.

PR #259, opened 2020-08-20
Avoid holding write transaction lock while accessing fork-join pool

erikrozendaal:
The write transaction wasn't needed in this case until all objects
were converted anyway.

Holding the write transaction lock could result in a deadlock if the
common fork-join pool is full of blocked threads (shouldn't happen...)
waiting for Xodus (read or write).

In general we should avoid accessing Xodus from the common fork-join
pool, this should also avoid this issue completely.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
@@ -361,9 +361,8 @@ public class RpkiRepositoryValidationService {
         AtomicInteger pendingObjectsBytes = new AtomicInteger(0);
         List<Pair<String, byte[]>> pendingObjects = new ArrayList<>(1000);
         Runnable commitPendingObjects = () -> {
-            storage.writeTx0((tx) -> counter.addAndGet(
-                    storePendingObjects(tx, validationRun, pendingObjects))
-            );
+            int count = storePendingObjects(validationRun, pendingObjects);
+            counter.addAndGet(count);
             pendingObjects.clear();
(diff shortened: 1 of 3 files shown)


lolepezy (approved):

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
```diff
@@ -429,15 +428,17 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
         commitPendingObjects.run();
     }
 
-    private int storePendingObjects(Tx.Write tx, RsyncRepositoryValidationRun validationRun, List<Pair<String,byte[]>> pendingObjects) {
+    private int storePendingObjects(RsyncRepositoryValidationRun validationRun, List<Pair<String,byte[]>> pendingObjects) {
         AtomicInteger counter = new AtomicInteger();
 
         // Parsing RPKI objects is CPU bound, so do this with any available threads
```
ties, line 434: I'd like if it we document somewhere that `parallelStream` while holding a transaction is a footcannon on machines with a low number of cores.

ties (approved):

wibisono (approved):

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
erikrozendaal: We should probably add some design notes documentation file with stuff like this. The bigger foot cannon (leg cannon?) is probably the other way around: running Xodus transactions from within the fork-join pool w/o managed blocking.

PR #260, opened 2020-08-20
Notify fork join pool of blocking operations

erikrozendaal:
Xodus transactions and HTTP downloads are (potentially) blocking. If
we accidentally call these from the fork join pool it may cause the
pool to be exhausted. By using the `ForkJoinPool#managedBlock` API
additional threads will be created if needed.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/xodus/Xodus.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/xodus/Xodus.java
@@ -43,6 +43,7 @@ import net.ripe.rpki.validator3.storage.data.Key;
 import net.ripe.rpki.validator3.storage.encoding.Coder;
 import net.ripe.rpki.validator3.storage.encoding.CoderFactory;
 import net.ripe.rpki.validator3.storage.Storage;
+import net.ripe.rpki.validator3.util.ForkJoin;
 import org.apache.commons.lang3.tuple.Pair;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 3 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/util/ForkJoin.java
```diff
@@ -0,0 +1,38 @@
...
+            ForkJoinPool.managedBlock(new ForkJoinPool.ManagedBlocker() {
+                @Override
+                public boolean block() {
+                    result.set(supplier.get());
+                    return true;
+                }
+
+                @Override
+                public boolean isReleasable() {
+                    return false;
```
wibisono, line 28: Blocking queue example I found [here](https://www.concretepage.com/java/jdk7/example-of-managedblocker-in-java) gives me impression that maybe here we can release if the result is set?

rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/xodus/Xodus.java
```diff
@@ -79,7 +80,7 @@ protected synchronized Store meta() {
         AtomicReference<List<Runnable>> afterCommitHooks = new AtomicReference<>(Collections.emptyList());
 
         Environment env = getEnv();
-        T result = env.computeInExclusiveTransaction(txn -> {
+        T result = ForkJoin.blocking(() -> env.computeInExclusiveTransaction(txn -> {
```
lolepezy, line 83: computeInExclusiveTransaction blocks the whole database, so there can only be one computeInExclusiveTransaction running at every moment. Does it change the logic?

lolepezy (approved):

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/ForkJoin.java
erikrozendaal: The fork-join pool will release the thread as soon as one of `isReleasable` or `block` returns true. So in our case it will call `isReleasable`,  get `false` so call `block`, then get `true` and then we're done.

rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/xodus/Xodus.java
erikrozendaal: There could be many threads waiting for the `computeInExclusiveTransaction` lock. All these threads should not be blocking the fork-join pool, so the logic here should be good.

ties (approved):

PR #262, opened 2020-08-25
Remove unused ROA prefixes from RpkiObject

erikrozendaal:
Remove unused ROA prefixes from RPKI object

These take up a lot of space and time and are not used. This also
solves the problem of storing a large ROA (1 MB) into Xodus with a
segment size of 8 MB (previously the 1 MB ROA expanded to 10 MB
encoded).

Decoding a 10 kB ROA dropped from ~300 ns to ~1.5 ns.


wibisono (approved):
Tested locally and loading as-0 works flawlessly, also export-extended still work.

rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/encoding/custom/RpkiObjectCoder.java
```diff
@@ -75,20 +65,14 @@
...
-        rpkiObject.setAuthorityKeyIdentifier(content.get(AKI_TAG));
-        Encoded.field(content, SIGNING_TIME_TAG).ifPresent(b -> rpkiObject.setSigningTime(Coders.toInstant(b)));
-        Encoded.field(content, SERIAL_TAG).ifPresent(b -> rpkiObject.setSerialNumber(Coders.toBigInteger(b)));
+        RpkiObject.Type type = RpkiObject.Type.valueOf(Coders.toString(content.get(TYPE_TAG)));
+        byte[] encoded = content.get(ENCODED_TAG);
 
-        Encoded.field(content, ROA_PREFIXES).ifPresent(b ->
-                rpkiObject.setRoaPrefixes(Coders.fromBytes(b, roaPrefixCoder::fromBytes)));
+        ValidationResult ignored = ValidationResult.withLocation("ignored." + type.getExtensions());
+        CertificateRepositoryObject certificateRepositoryObject = CertificateRepositoryObjectFactory.createCertificateRepositoryObject(encoded, ignored);
```
lolepezy, line 72: I remember that the whole point of having encoding different from original ASN1 was the speed -- it's way slower to extract stuff from ASN1 compared to serialisation libraries (or this implementation). Otherwise, we would simply kept original binaries and wouldn't bother with any serialisation at all. 

Can you benchmark how does this influence the deserialisation speed? Because we already consume a lot of CPU on tree validation and this change can make it even worse.

lolepezy (changes requested):

rpki-validator/src/main/java/net/ripe/rpki/validator3/storage/encoding/custom/RpkiObjectCoder.java
lolepezy: Essentially we need to look at how much CPU-time it take to parse all the objects in a rsync/rrdp repository and add it on top of every tree validation. AFAICR that was quite some overhead.

erikrozendaal: (Unscientific) benchmark with a 10 kB ROA (from `src/test/resources`) indicates ASN.1 parsing is about 3 times slower:

Using `RoaPrefixCodec`:

```
Run 0 took 3551ms (355.1 ns per object)
Run 1 took 3169ms (316.90000000000003 ns per object)
Run 2 took 3054ms (305.4 ns per object)
Run 3 took 3074ms (307.40000000000003 ns per object)
Run 4 took 3090ms (309.0 ns per object)
Run 5 took 3100ms (310.0 ns per object)
Run 6 took 3062ms (306.20000000000005 ns per object)
Run 7 took 3048ms (304.8 ns per object)
Run 8 took 3085ms (308.5 ns per object)
Run 9 took 3076ms (307.59999999999997 ns per object)
```

Using ASN.1 decoding:

```
Run 0 took 10118ms (1011.8000000000001 ns per object)
Run 1 took 8668ms (866.8000000000001 ns per object)
Run 2 took 8493ms (849.3000000000001 ns per object)
Run 3 took 8548ms (854.8 ns per object)
Run 4 took 8925ms (892.5 ns per object)
Run 5 took 8619ms (861.9 ns per object)
Run 6 took 8582ms (858.1999999999999 ns per object)
Run 7 took 8650ms (865.0 ns per object)
Run 8 took 8526ms (852.6 ns per object)
Run 9 took 8578ms (857.8 ns per object)
```

However, it turns out that the ROA prefixes from `RpkiObject` are not used at all by the application. Then it gets really fast:

```
Run 0 took 36ms (3.6 ns per object)
Run 1 took 14ms (1.4 ns per object)
Run 2 took 20ms (2.0 ns per object)
Run 3 took 21ms (2.1 ns per object)
Run 4 took 15ms (1.5 ns per object)
Run 5 took 20ms (2.0 ns per object)
Run 6 took 19ms (1.9 ns per object)
Run 7 took 17ms (1.7 ns per object)
Run 8 took 16ms (1.6 ns per object)
Run 9 took 19ms (1.9 ns per object)
```

wibisono (approved):

lolepezy (approved):
Removing code is the best!

PR #263, opened 2020-08-26
RPKI repository cleanup job

erikrozendaal:
Track the time that a repository was last referenced during a validation run. Once a repository hasn't been referenced for at least `rpki.validator.rpki.repository.cleanup.grace.duration` it will be deleted. Deleted RRDP repositories will also be removed from the validation scheduler.

Best way to review is probably by going through each commit separately.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/trustanchors/TrustAnchorService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/api/trustanchors/TrustAnchorService.java
@@ -36,8 +36,6 @@ import net.ripe.rpki.validator3.domain.validation.TrustAnchorState;
 import net.ripe.rpki.validator3.domain.validation.ValidatedRpkiObjects;
 import net.ripe.rpki.validator3.storage.Tx;
 import net.ripe.rpki.validator3.storage.data.Key;
-import net.ripe.rpki.validator3.storage.data.Ref;
-import net.ripe.rpki.validator3.storage.data.RpkiRepository;
 import net.ripe.rpki.validator3.storage.data.TrustAnchor;
 import net.ripe.rpki.validator3.storage.Storage;
 import net.ripe.rpki.validator3.storage.stores.RpkiRepositories;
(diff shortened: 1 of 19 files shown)


lolepezy:
Why do we need per-TA timestamp? Wouldn't it be enough to keep one only timestamp as the earliest of "refereced by TA"?

erikrozendaal:
> Why do we need per-TA timestamp? Wouldn't it be enough to keep one only timestamp as the earliest of "refereced by TA"?

When a TA is added to a repository it will always be re-validated after the repository updates. So we should remove a TA that no longer references a repository, even when other TA's still do. Probably very much a corner case.

ties (approved):
Looks good (looked at the separate commits). One nit: It may be good to squash the commit that is not forward-compatible w.r.t. the database before morning, because it would be bad to land on that one during a bisect.

lolepezy:
> > Why do we need per-TA timestamp? Wouldn't it be enough to keep one only timestamp as the earliest of "refereced by TA"?
> 
> When a TA is added to a repository it will always be re-validated after the repository updates. So we should remove a TA that no longer references a repository, even when other TA's still do. Probably very much a corner case.

Ehm, I still don't get it. When a repository is referred (re-registred) in some tree validation process its `lastReferencedAt` gets bumped to the current time. I would think that it doesn't matter from which TAs it happens, how many of them is there, etc. It would probably need some explanatory comment why do we need multiple timestamps.

erikrozendaal:
I've updated the comment on `RpkiRepository#trustAnchors` to

```
The set of trust anchors that referenced this repository during validation. Whenever this RPKI repository
is updated re-validation for this set of trust anchors will be triggered. Normally a repository is only
referenced by a single trust anchor, but this is not guaranteed.

For each trust anchor we store the last time the trust anchor referenced this repository during validation.
When a trust anchor no longer references a this repository during validation it will be removed from this
set. This way the trust anchor will no longer be re-validated when this repository is updated.

When the repository is no longer referenced by any trust anchor it will be removed and no longer updated
periodically.
```

Does this explain it correctly? We just do not want to trigger validation for trust anchors that no longer reference the repository, even when it is still referenced by another trust anchor.

lolepezy:
> 
> Does this explain it correctly? We just do not want to trigger validation for trust anchors that no longer reference the repository, even when it is still referenced by another trust anchor.

Thanks, now it's much more clear.

lolepezy (approved):

PR #265, opened 2020-08-31
Fix error time reporting

lolepezy:
This is to 
a) Fix 'Seen at' so that it show UTC time of when the error was seen
b) Fix the 'strict' mode using rpki-commons-1.13 so that we don't have 'paranoid' mode (complain about objects 7 hours before they expire).

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/config/ValidationConfig.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/config/ValidationConfig.java
@@ -43,14 +43,26 @@ public class ValidationConfig {
     @Setter
     private boolean strictValidation;
 
+    /**
+     * Internally used option: strict validation with _negative_ grace periods,
+     * i.e. validator starts to complain about objects before they expire. This
+     * options is supposed to be used only for internal RIPE NCC repository testing.
+     */
+    @Getter
+    @Setter
+    private boolean paranoidValidation = false;
+
     @Getter
     @Setter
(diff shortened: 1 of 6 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/config/ValidationConfig.java
```diff
@@ -39,16 +39,30 @@
 @ConfigurationProperties("rpki.validator")
 public class ValidationConfig {
 
-    @Getter @Setter
+    @Getter
+    @Setter
     private boolean strictValidation;
 
-    @Getter @Setter
+    /**
```
ties, line 46: Perhaps this name isn't very good.
`earlyWarningValidation`?

ui-validator/src/assets/i18n/en.json
```diff
@@ -24,7 +24,7 @@
   "OF": "of",
   "SHOW": "Show",
   "ENTRIES": "entries",
-  "SEEN_AT": "Seen at",
+  "SEEN_AT": "Seen at (UTC)",
```
ties, line 27: Good for consistency to use UTC instead of local.

ties (approved):

rpki-validator/src/main/java/net/ripe/rpki/validator3/config/ValidationConfig.java
lolepezy: Yeah, makes sense.

PR #268, opened 2020-09-02
Maximize managed blocker threads

erikrozendaal:
Ensures we no longer take up all the spare threads in the common fork-join pool that can cause `RejectedExecutionExceptions`. 

I'm not sure of all this is worth the trouble. The validator worked before without any `blocking` implementation so an alternative is to just remove this whole concept for simplicity.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
@@ -337,9 +337,9 @@ public class CertificateTreeValidationService {
             }
             accumulator.add(manifestObject.get().key(), manifest, ImmutableSortedSet.of(temporary.getCurrentLocation().getName()));
 
-            manifest.getFiles().entrySet()
-                .parallelStream()
+            manifest.getFiles().entrySet().stream()
                 .flatMap(entry -> getManifestEntry(manifestUri, entry))
+                .parallel()
                 .flatMap(tuple -> getCertificateRepositoryObjectValidationContext(trustAnchor, context, accumulator, crlUri, x509Crl, tuple))
                 .map(tuple -> {
                     final ValidationResult vr = tuple.v2();
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/util/ForkJoin.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/util/ForkJoin.java
@@ -42,9 +42,14 @@ public class ForkJoin {
     private static final Semaphore maximumManagedBlockers;
 
     static {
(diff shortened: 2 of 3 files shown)


ties (approved):
I think the code itself is clear.

However I agree that it is complexity that we may no longer need now that we solved the scenario causing the deadlock.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -337,9 +337,9 @@ private void validateCertificateAuthority(final TrustAnchor trustAnchor,
             }
             accumulator.add(manifestObject.get().key(), manifest, ImmutableSortedSet.of(temporary.getCurrentLocation().getName()));
 
-            manifest.getFiles().entrySet()
-                .parallelStream()
+            manifest.getFiles().entrySet().stream()
                 .flatMap(entry -> getManifestEntry(manifestUri, entry))
+                .parallel()
```
lolepezy, line 342: Why this change? Not arguing, but just out of curiosity.

rpki-validator/src/main/java/net/ripe/rpki/validator3/util/ForkJoin.java
```diff
@@ -29,22 +29,47 @@
...
 import java.util.function.Supplier;
 
+@Slf4j
 public class ForkJoin {
+
+    private static final Semaphore maximumManagedBlockers;
+
+    static {
+        int maximumSpares = Integer.parseInt(System.getProperty("java.util.concurrent.ForkJoinPool.common.maximumSpares", "256"));
+        int max = Math.min(2 * ForkJoinPool.getCommonPoolParallelism(), maximumSpares / 4);
```
lolepezy, line 46: Why maximumSpares / 4? 
I would want a comment with a link to some reasoning about these choices.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
erikrozendaal: From the commit message:

```
  Load manifest entries sequentially but validate concurrently

Loading manifest entries is mostly IO bound (Xodus read transactions)
so there is no need for parallel processing. Actual validation is
still in parallel since it is CPU intensive.
```

Still, now that `blocking` is removed from the Xodus `readTx` it probably doesn't matter much anyway.

wibisono (approved):
Rather reluctant to maintain the complexity, but it seems that it is better to have it. At least next time we perform writeTx within parallel stream we wouldn't have deadlock.

lolepezy (approved):

PR #270, opened 2020-09-09
Fail manifest validation if any listed object fails to validate

erikrozendaal:
Supports better strict mode checking as defined by the RFC 6486bis (section 6, https://tools.ietf.org/html/draft-ymbk-sidrops-6486bis-00#section-6).

If any object on the manifest fails validation the manifest is considered invalid and no ROAs, router certificate, or child CAs will be processed.

Other checks (such as file name checking, issuing a warning, falling back to previous results, etc) are not part of this PR.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
@@ -30,7 +30,7 @@
 package net.ripe.rpki.validator3.domain.validation;
 
 import com.google.common.collect.ImmutableSortedSet;
-import lombok.Getter;
+import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
 import net.ripe.rpki.commons.crypto.CertificateRepositoryObject;
 import net.ripe.rpki.commons.crypto.cms.manifest.ManifestCms;
@@ -65,12 +65,12 @@ import net.ripe.rpki.validator3.util.Bench;
 import net.ripe.rpki.validator3.util.Time;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.jooq.lambda.tuple.Tuple2;
-import org.jooq.lambda.tuple.Tuple3;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import java.net.URI;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -255,170 +255,178 @@ private boolean isValidationRunCompleted(ValidationResult validationResult) {
...
+            TrustAnchor trustAnchor,
+            CertificateRepositoryObjectValidationContext context,
+            URI manifestUri,
+            RpkiObject manifestObject,
+            ManifestCms manifest,
+            URI crlUri,
+            X509Crl x509Crl,
+            Accumulator accumulator
+    ) {
+        ValidationResult temporary = newValidationResult(context.getLocation());
```
wibisono, line 359: Personal preference, variable rename into `result` in the `validateCertificateAuthority` above makes it clear that it will be the result rather than `temporary`

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -255,170 +255,178 @@ private boolean isValidationRunCompleted(ValidationResult validationResult) {
...
+        List<ManifestEntryValidationResult> objectValidationResult = manifest.getFiles().entrySet().parallelStream()
+                .map(entry -> validateManifestEntry(trustAnchor, context, manifestUri, crlUri, x509Crl, entry))
+                .collect(toList());
+
+        objectValidationResult.stream().map(ManifestEntryValidationResult::getValidationResult).forEach(temporary::addAll);
```
wibisono, line 364: Maybe we can immediately add validation result while parallel mapping above? Or is there a reason to add it later/separately here.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -255,170 +255,178 @@ private boolean isValidationRunCompleted(ValidationResult validationResult) {
...
+        List<ValidManifestObject> validObjects = objectValidationResult.stream().flatMap(ValidManifestObject::of).collect(toList());
+
+        accumulator.add(manifestObject.key(), manifest, ImmutableSortedSet.of(manifestUri.toASCIIString()));
+        validObjects.forEach(object -> accumulator.add(
+                object.key(),
+                object.getCertificateRepositoryObject(),
+                ImmutableSortedSet.of(object.getLocation().toASCIIString()))
+        );
+
+        List<CertificateRepositoryObjectValidationContext> children = validObjects.stream().flatMap(object -> {
```
lolepezy, line 378: This one can potentially get pretty big (we have a big flat manifest of the RIPE NCC's production CA). Maybe not a big deal, but if we can avoid instantiating the whole list in memory, would be nice.

lolepezy (approved):

lolepezy:
Looks good to me.

wibisono (approved):

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
erikrozendaal: We'd need to perform additional locking since `ValidationResult` is not thread-safe. I also prefer methods with fewer side-effects.

PR #272, opened 2020-09-16
Use depth-first CA validation to reduce memory usage

erikrozendaal:
The code relies a bit less on side-effects and the validation
functions now return a result instead of mutating an accumulator.

All certificate authorities are validated, even when a manifest
contains errors in strict mode. However, the results of the validation
are not kept. This way we avoid having keep child certificates in
memory or to parse the manifest entries twice (first to discover
errors in the manifest, then to process the children).

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
@@ -48,6 +48,7 @@ import net.ripe.rpki.validator3.config.ValidationConfig;
 import net.ripe.rpki.validator3.domain.ErrorCodes;
 import net.ripe.rpki.validator3.domain.metrics.TrustAnchorMetricsService;
 import net.ripe.rpki.validator3.domain.validation.ValidatedRpkiObjects.Accumulator;
+import net.ripe.rpki.validator3.domain.validation.ValidatedRpkiObjects.TrustAnchorData;
 import net.ripe.rpki.validator3.storage.Storage;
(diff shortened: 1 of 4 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -255,189 +258,185 @@ private boolean isValidationRunCompleted(ValidationResult validationResult) {
...
+                .collect(
+                        CertificateAuthorityValidationResult::empty,
+                        CertificateAuthorityValidationResult::addAll,
+                        CertificateAuthorityValidationResult::addAll
+                );
+
+        if (validationConfig.isStrictValidation() && result.getValidationResult().hasFailures()) {
+            // RFC 6486bis: drop all validated objects from this CA (and its children) since there
+            // was at least one validation failure on this manifest.
+            return CertificateAuthorityValidationResult.of(result.getValidationResult());
```
erikrozendaal, line 370: Should we also include the validation results from the (dropped) child certificate authorities here? In this case we need to use `result.getAllValidationResults()`.

ties (approved):

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
lolepezy: I agree that it will create a lot of confusion to see errors about objects that are not really taken into account.

lolepezy (approved):

wibisono (approved):

PR #280, opened 2020-09-23
Existing RPKI repository "referenced at" is not updated by validation

erikrozendaal:
All existing RPKI repositories for the trust anchor are pre-loaded
when tree validation starts. This means they are no longer
`registered` during the validation process and their last referenced
at timestamp is not updated. At some point the repository cleanup job
will then delete these repositories from the store.

This causes tree validation to no longer start (since it is only
triggered by a repository update), resulting in the data for the trust
anchor to be no longer updated.

Fixed by no longer pre-loading the repositories and let the regular
validation process register the repositories as needed. Any
repositories that are no longer referenced will be deleted at some
point.


lolepezy (approved):

ties (approved):

wibisono (approved):

wibisono:
LGTM, just wondering if this changes is expected to have an impact on bootstrapping.

PR #285, opened 2020-10-07
Case insensitive URI schemes

erikrozendaal:
File system names and indexed location names are not handled in a case insensitive manner as this probably opens up a whole new can of worms when running on case insensitive file systems, etc. So the same is true for the URI index in `RpkiRepositoriesStore`.

### DIFF
--- a/rpki-parent/pom.xml
+++ b/rpki-parent/pom.xml
@@ -18,7 +18,7 @@
 
         <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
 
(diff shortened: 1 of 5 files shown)


rpki-parent/pom.xml
```diff
@@ -18,7 +18,7 @@
 
         <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
 
-        <rpki-commons.version>1.13</rpki-commons.version>
```
ties, line 21: Are we at a point to make a release of that lib instead of snapshot?

ties (approved):

wibisono (approved):

PR #287, opened 2020-10-13
Decrease bootstrap time to first validation

erikrozendaal:
Fixes two causes of long delays in the bootstrapping process:

1. Triggering already running certificate tree validations were ignored. So if a repository was updated while a validation is already running it could take 10+ minutes before the certificate tree validation was run again, based on the frequency of repository updates. Now the trigger is remembered and the validation will be scheduled again after the current run completes. The minimum time between validation runs was also reduced, as it is mainly used to avoid quickly triggering multiple runs.

2. The database was only checked once every 5 minutes for PENDING rsync repositories. Now we check every minute.

Together these changes result in a bootstrap time between 3-4 minutes (also depending on network speed, etc) using the 5 RIR trust anchors and the AS0 trust anchor.

Note that the `Throttled` class requires careful review since it needs to be concurrency aware and safe.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
@@ -63,6 +63,11 @@ public class Throttled<Key> {
 
     private final long minIntervalMs;
 
+    /*
+     * Tracks the action for each Key. This map is used concurrently by multiple threads
+     * so must be locked for any access. This might become a bottleneck when there are many
+     * keyed tasks, but currently this is not the case.
+     */
     private final Map<Key, Action> actionMap = new HashMap<>();
 
     private final ScheduledExecutorService scheduledExecutor =
@@ -78,13 +83,23 @@ public class Throttled<Key> {
             final Action action = actionMap.get(key);
 
             if (action == null) {
+                // New key, start running action as soon as possible.
                 actionMap.put(key, Action.toBeExecutedASAP(now, r));
                 scheduledExecutor.execute(actionRunner(key));
             } else if (action.running) {
+                // Action for this key is currently running, mark the
+                // action as scheduled so that it will be started again
+                // as soon as the current one finishes.
                 actionMap.put(key, action.scheduled(r));
             } else if (action.alreadyScheduled) {
+                // Action already scheduled for this key, only replace
+                // the task that needs to be run when it starts.
                 actionMap.put(key, action.replaced(r));
             } else {
+                // Action not running or scheduled, so schedule it for
+                // execution taking into account the minIntervalMs.
+                // Note that a negative delay will schedule the action for
+                // immediate execution.
                 final Instant lastTime = action.getExecutionTime();
                 final long delay = minIntervalMs - (now.toEpochMilli() - lastTime.toEpochMilli());
                 actionMap.put(key, action.scheduled(r));
@@ -105,7 +120,7 @@ public class Throttled<Key> {
             } finally {
                 synchronized (actionMap) {
                     Action action = actionMap.get(key);
-                    actionMap.put(key, action.stopped());
+                    actionMap.put(key, action.finished());
                     if (action.alreadyScheduled) {
                         trigger(key, action.runnable);
                     }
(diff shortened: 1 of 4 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
```diff
@@ -44,85 +43,104 @@
...
+            if (action == null) {
+                actionMap.put(key, Action.toBeExecutedASAP(now, r));
+                scheduledExecutor.execute(actionRunner(key));
+            } else if (action.running) {
+                actionMap.put(key, action.scheduled(r));
+            } else if (action.alreadyScheduled) {
+                actionMap.put(key, action.replaced(r));
+            } else {
+                final Instant lastTime = action.getExecutionTime();
+                final long delay = minIntervalMs - (now.toEpochMilli() - lastTime.toEpochMilli());
```
lolepezy, line 89: Should we handle negative delay case? Can it ever happen?

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
```diff
@@ -44,85 +43,104 @@
...
-                        actionMap.put(key, action.scheduled());
-                        scheduledExecutor.schedule(wrappedRunnable, delay, TimeUnit.SECONDS);
-                    } else {
-                        actionMap.put(key, action.scheduled());
-                        scheduledExecutor.execute(wrappedRunnable);
+                action.runnable.run();
+            } finally {
+                synchronized (actionMap) {
+                    Action action = actionMap.get(key);
+                    actionMap.put(key, action.stopped());
```
lolepezy, line 108: What is "stopped" for?

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
```diff
@@ -44,85 +43,104 @@
...
+                    if (action.alreadyScheduled) {
+                        trigger(key, action.runnable);
```
lolepezy, line 110: And why do we trigger action for the same key again? To process one if it was waiting on the `synchronized`?

ties (commented):
Can you write down (preferably in block comments) what invariants you expect to hold in the `actionMap`?

It's hard for me to reason about correctness right now.

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
erikrozendaal: Negative delay results in an immediate execution.

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
erikrozendaal: Renamed to `finished` to indicate that the current execution is finished.

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
erikrozendaal: To process actions that were scheduled while the action was running. So we wait until the running action finishes to re-trigger the action, taking into account the `minIntervalMs`.

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
```diff
@@ -44,85 +43,104 @@
...
+ * Thread.sleep(11000)
+ * <p>
+ * throttled.trigger("action1", r5) - will execute r5 immediately again.
  */
 public class Throttled<Key> {
 
-    private final long minIntervalInSeconds;
+    private final long minIntervalMs;
 
     private final Map<Key, Action> actionMap = new HashMap<>();
```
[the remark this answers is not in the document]
erikrozendaal: Not easily. We often take decisions based on the current value and perform side-effects (schedule execution). `ConcurrentHashMap` will often call your map's `compute` methods multiple times, so side-effects may also run multiple times.

We could try to lock on the `action` instance for higher concurrency, but then this class can no longer be immutable which is a bigger refactoring than the one here.

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/Throttled.java
```diff
@@ -44,85 +43,122 @@
  * The purpose of this class is to be able to execute an arbitrary action
```
wibisono, line 43: This might makes it more obvious to library user: " to execute an arbitrary action having the same action key not more often than a certain interval", and in the example illustrate that "action1" is actually "action1Key".

wibisono (approved):

lolepezy (approved):

PR #289, opened 2020-10-15
Revalidate certificate tree when objects expire

erikrozendaal:
Tracks and shows the earliest object expiration time per trust anchor:

 image

Re-validates the trust anchor certificate tree when this time is in the past.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
@@ -210,8 +210,10 @@ public class CertificateTreeValidationService {
                 }
             }
 
-            log.info("TA {} next validation time {}", trustAnchor.getName(), accumulator.getNextUpdateTime());
-            validationRun.setNextValidationNeededAt(InstantWithoutNanos.from(accumulator.getNextUpdateTime()));
+            if (accumulator.getNextUpdateTime() != null) {
+                log.info("TA {} next validation time {}", trustAnchor.getName(), accumulator.getNextUpdateTime());
+                validationRun.setNextValidationNeededAt(InstantWithoutNanos.from(accumulator.getNextUpdateTime()));
+            }
 
             storage.writeTx0(tx -> {
                 validationRuns.add(tx, validationRun);
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
@@ -210,6 +210,9 @@ public class CertificateTreeValidationService {
                 }
             }
 
+            log.info("TA {} next validation time {}", trustAnchor.getName(), accumulator.getNextUpdateTime());
+            validationRun.setNextValidationNeededAt(InstantWithoutNanos.from(accumulator.getNextUpdateTime()));
+
             storage.writeTx0(tx -> {
                 validationRuns.add(tx, validationRun);
                 Long t = Time.timed(() -> accumulator.forEach(key -> validationRuns.associateRpkiObjectKey(tx, validationRun, key)));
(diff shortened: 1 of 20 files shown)


lolepezy:
It looks a little confusing: any user would think "AfriNic is going to revalidate tomorrow" and that's very different from what is set by all these "*interval*" settings in the config.

Do I understand correctly that it's actually the time when TA _has_ to be validated again because objects will start to expire?

lolepezy:
One thing I would want to make sure not going to happen -- if some objects are actually expired, are we going to spin in the loop revalidating the TA all over again?

lolepezy:
And generally, what is the problem this PR solves?

erikrozendaal:
> It looks a little confusing: any user would think "AfriNic is going to revalidate tomorrow" and that's very different from what is set by all these "_interval_" settings in the config.
> 
> Do I understand correctly that it's actually the time when TA _has_ to be validated again because objects will start to expire?

Yes. Maybe "earliest object expiration"? Valid until? Not valid after?

erikrozendaal:
> One thing I would want to make sure not going to happen -- if some objects are actually expired, are we going to spin in the loop revalidating the TA all over again?

Expired objects are invalid and are not taken into account when determining the `nextValidatedNeededAt` time. Certification is also throttled. But it could happen that a lot of objects expire one after another resulting in multiple runs, at most one per minute per TA.

erikrozendaal:
> And generally, what is the problem this PR solves?

Currently validated ROA prefixes and router certificates remain valid forever when certificate tree validation isn't run. This could happen when the CA no longer publishes objects. With this process expired objects will no longer be part of the the validation results.

lolepezy:
> > And generally, what is the problem this PR solves?
> 
> Currently validated ROA prefixes and router certificates remain valid forever when certificate tree validation isn't run. This could happen when the CA no longer publishes objects. With this process expired objects will no longer be part of the the validation results.

Would it help it revalidate TA unconditionally every `max(rsync-intrerval, rrdp-interval)`? Or something along the lines "do the tree validation every N minutes unless it's already validated less than N minutes ago".

Why I'm being picky about it: this PR is introducing the dependency of the running schedule from the data coming from somewhere on the internets, thus opening the can of worms of "it's spinning my CPU and I have not idea why"-type of bugs. And since we are going to kill it soon anyway, I would prefer it to be dumb and simple and not employ fancy scheduling strategies.

In principle, the change seems to be fine, except that it's hard to say what can potentially go wrong here.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/ValidatedRpkiObjects.java
```diff
@@ -292,6 +299,28 @@ public void add(TrustAnchorData trustAnchorData, Key key, CertificateRepositoryO
                         pkInfo
                 ));
             }
+
+            DateTime expiresAt = null;
+            if (object instanceof RpkiSignedObject) {
+                RpkiSignedObject obj = (RpkiSignedObject) object;
+                expiresAt = obj.getNotValidAfter();
```
lolepezy, line 306: Do we need to check both expiration time of the EE certificate and CMS-specific 'notValidaAfter' here? According to ROS report they are different on out manifests, so it would make sense to double-check here.

lolepezy:
> 
> Yes. Maybe "earliest object expiration"? Valid until? Not valid after?

'Earliest object expiration' sounds good to me, yes.

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/ValidatedRpkiObjects.java
erikrozendaal: Fixed

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -210,6 +210,11 @@ private void validateTa(TrustAnchor trustAnchor) {
                 }
             }
 
+            if (accumulator.getNextUpdateTime() != null) {
```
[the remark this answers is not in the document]
erikrozendaal: Next update time can be the same when the validation run was triggered by a repository update or something like that. When triggered by the `CertificateTreeObjectExpirationValidationRun` the old next update time is always in the past. So the updated next update time will either be null (no valid objects) or somewhere in the future, never equal to the previous one.

erikrozendaal:
> Why I'm being picky about it: this PR is introducing the dependency of the running schedule on the data coming from somewhere on the internets, thus opening the can of worms of "it's spinning my CPU and I have not idea why"-type of bugs. And since we are going to kill it soon anyway, I would prefer it to be dumb and simple and not employ fancy scheduling strategies.

If there are no more updates than validation jobs will be scheduled at most once a minute (for each trust anchor). However, this also requires many objects with "not valid after" times just one minute apart. And at some point it will end when all objects have expired...

I wanted to avoid running tree validation all the time just in case an object has expired, that's why this job is only triggered when needed.

Normally repository updates will trigger validation long before object expiration happens.

ties (approved):

PR #294, opened 2020-11-16
update to rpki-commons 1.15

ties:


ties:
Confirming that `docker.io/ripencc/rpki-validator-3:beta` runs, merging and creating an RC.

PR #297, opened 2020-11-24
Fix bgp update time

lolepezy:
Fix DownloadBgpRisDumpsJob not running on time.


rpki-validator/src/main/java/net/ripe/rpki/validator3/background/BackgroundJobs.java
```diff
@@ -51,72 +46,71 @@
...
+            jobToBeExecuted(key);
+            try {
+                r.run();
+            } catch (Exception e) {
+                log.error(String.format("Error executing job '%s'", key), e);
+            } finally {
+                jobWasExecuted(key);
+            }
+        };
+        scheduledExecutorService.scheduleAtFixedRate(rr, delay, period, timeUnit);
```
erikrozendaal, line 112: Maybe `scheduleWithFixedDelay` to avoid running jobs concurrently when they take a long time to run?

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/BackgroundJobs.java
```diff
@@ -51,72 +46,71 @@
...
+    private final BgpPreviewService bgpPreviewService;
 
-        scheduler.getListenerManager().addJobListener(this);
+    @Autowired
+    public BackgroundJobs(RpkiObjectCleanupService rpkiObjectCleanupService,
+                          RpkiRepositoryCleanupService rpkiRepositoryCleanupService,
+                          ValidationRunCleanupService validationRunCleanupService,
+                          RpkiRepositoryValidationService rpkiRepositoryValidationService,
+                          BgpPreviewService bgpPreviewService) {
+        this.scheduledExecutorService = Executors.newScheduledThreadPool(Runtime.getRuntime().availableProcessors());
```
erikrozendaal, line 74: Quartz uses a fixed pool of 10 threads by default. The `newScheduledThreadPool` has no maximum by default. We could just use `0` as the core pool size and let the thread pool create threads as needed, there should be little overhead.

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/BackgroundJobs.java
lolepezy: Documentation says it will never happen.

rpki-validator/src/main/java/net/ripe/rpki/validator3/background/BackgroundJobs.java
lolepezy: Indeed!
From what I can see there's not way to create ScheduledThreadPoolExecutor with a cap on the thread pool size. No idea why actually, to always have a thread to submit to? Will make it 0 then.

PR #301, opened 2020-12-08
Only fail the entire manifest when entry hashes do not match

erikrozendaal:
If included objects fail validation, only drop the failed objects, not
all objects on the manifest. The main intent is to prevent MITM
attacks where an attacker replaces or removes objects from the
publication point. Checking that all manifest entries are present and
the hashes match prevents this.

Bugs in the CA or publication server that cause invalid objects to be
published (with the correct hash on the manifest) will not be
prevented and will not cause all objects to become invalid.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
@@ -63,6 +63,7 @@ import net.ripe.rpki.validator3.storage.stores.Settings;
 import net.ripe.rpki.validator3.storage.stores.TrustAnchors;
 import net.ripe.rpki.validator3.storage.stores.ValidationRuns;
 import net.ripe.rpki.validator3.util.Bench;
+import net.ripe.rpki.validator3.util.Sha256;
 import net.ripe.rpki.validator3.util.Time;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -377,15 +378,15 @@ public class CertificateTreeValidationService {
 
         if (!validationConfig.isStrictValidation()) {
             return result;
-
         }
 
-        boolean hasFailures = result.getValidationResult().hasFailures();
+        // If any entry was not present or the hash did not match fail the entire manifest
         result.getValidationResult().setLocation(new ValidationLocation(manifestUri));
-        result.getValidationResult().warnIfTrue(hasFailures, MANIFEST_ALL_ENTRIES_VALID);
-        if (hasFailures) {
+        boolean hasManifestEntryFailures = result.getValidationResult().hasFailureForCurrentLocation();
+        result.getValidationResult().warnIfTrue(hasManifestEntryFailures, MANIFEST_ALL_ENTRIES_VALID);
+        if (hasManifestEntryFailures) {
             // RFC 6486bis: drop all validated objects from this CA (and its children) since there
-            // was at least one validation failure on this manifest.
+            // was at least one manifest entry validation failure on this manifest.
             return CertificateAuthorityValidationResult.of(result.getValidationResult());
         } else {
             return result;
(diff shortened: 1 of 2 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -401,31 +402,35 @@ private CertificateAuthorityValidationResult validateManifestEntry(
...
     ) {
-        Key key = Key.of(entry.getValue());
-        URI location = manifestUri.resolve(entry.getKey());
-        ValidationResult validations = newValidationResult(location);
+        ValidationResult validations = newValidationResult(manifestUri);
         CertificateAuthorityValidationResult result = CertificateAuthorityValidationResult.of(validations);
 
-        Optional<RpkiObject> object = storage.readTx(tx -> rpkiObjects.findBySha256(tx, entry.getValue()));
-        validations.rejectIfFalse(object.isPresent(), VALIDATOR_MANIFEST_ENTRY_FOUND, manifestUri.toASCIIString());
+        String entryFilename = entry.getKey();
```
ties, line 408: 👍 (nit: final?)

ties (approved):
Nice! I'm glad that the new intent/interpretation of `-bis` w.r.t. valid objects moves us to a much better spot in the design-space.

wibisono (approved):

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -401,31 +402,35 @@ private CertificateAuthorityValidationResult validateManifestEntry(
...
-        validations.rejectIfFalse(hashMatches, VALIDATOR_MANIFEST_ENTRY_HASH_MATCHES, entry.getKey());
+        boolean hashMatches = Arrays.equals(Sha256.hash(rpkiObject.getEncoded()), entryHash);
+        validations.rejectIfFalse(hashMatches, VALIDATOR_MANIFEST_ENTRY_HASH_MATCHES, entryFilename);
         if (validations.hasFailureForCurrentLocation()) {
             return result;
         }
 
         SortedSet<String> locations = storage.readTx(tx -> rpkiObjects.getLocations(tx, rpkiObject.key()));
-        validations.rejectIfFalse(locations.contains(location.toASCIIString()), VALIDATOR_REPOSITORY_NOT_AT_EXPECTED_LOCATION, location.toASCIIString());
+        validations.rejectIfFalse(locations.contains(entryLocation.toASCIIString()), VALIDATOR_REPOSITORY_NOT_AT_EXPECTED_LOCATION, entryLocation.toASCIIString());
```
wibisono, line 426: Looks ok,  just trying to understand better. If object is present based on check on line 413, can we still have a case where `entryLocation` is not on list of `locations`?

erikrozendaal: In theory the object could be published in some other repository. Since we perform all our lookups by SHA-256 hash we do  not necessarily know where the object was located. This check ensures it was located at the expected location.

lolepezy:
@erikrozendaal Could you point to a specific point in the code that transforms the logic from "MFT is valid when ALL objects on it are valid" back to the logic "MFT is valid when AT LEAST ONE object on it is valid"? Because, that's was the point of the latest SIDROPS decision, IIUC.

erikrozendaal:
> @erikrozendaal Could you point to a specific point in the code that transforms the logic from "MFT is valid when ALL objects on it are valid" back to the logic "MFT is valid when AT LEAST ONE object on it is valid"? Because, that's was the point of the latest SIDROPS decision, IIUC.

The change is that before we would fail an entire manifest if any object had any validation failure. Now we only check for validation failures related to the manifest itself (https://github.com/RIPE-NCC/rpki-validator-3/pull/301/files#diff-18eb933c19d23ed3982b15fc94b1000fa4e53d9eedacb49716ad93311131b799R385). Manifest entry checks (filename, SHA-256 hash) are now reported against the manifest.

This is based on the sidrops mailings:

> The objects ON the manifest must be present and matching hashes, and the manifest's EE certificate being correct, non-expired. I think this is what is meant with the manifest must be 'valid', and 'current'

So the validity of the objects on the manifest is not important to the validity of the manifest itself. Only that all objects mentioned are present and the hash matches.

PR #302, opened 2020-12-08
Validate CRL and manifest next update times are the same

erikrozendaal:
Or at least close: if they are within 60 seconds we validate. Otherwise we warn. Once CAs fix their code we can become more strict.

### DIFF
--- a/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
+++ b/rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
@@ -95,6 +95,8 @@ import static net.ripe.rpki.validator3.storage.data.RpkiRepository.Type.RSYNC;
 @Service
 @Slf4j
 public class CertificateTreeValidationService {
+    public static final int MAX_NEXT_UPDATE_DIFFERENCE_MS = 60_000;
+
     public final long LONG_DURATION_WARNING_MS = 60_000;
 
     private final ValidationConfig validationConfig;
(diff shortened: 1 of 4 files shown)


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
```diff
@@ -341,6 +343,9 @@ private CertificateAuthorityValidationResult validateCertificateAuthority(
                 return result;
             }
 
+            long nextUpdateTimeDifferenceMs = manifest.getNextUpdateTime().getMillis() - x509Crl.getNextUpdateTime().getMillis();
```
ties, line 346: Where does the requirement come from that they have the same timestamp?

I feel like warning earlier & rejecting (with a higher threshold) may be the right thing to do here. But that would require us to run this for a while.

ties (approved):

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/CertificateTreeValidationService.java
erikrozendaal: https://tools.ietf.org/html/draft-ietf-sidrops-6486bis-03#section-4.2.1 under `nextUpdate`:

> If a manifest encompasses a CRL, the nextUpdate field of the manifest MUST match that of the CRL's nextUpdate field, as the manifest will be re-issued when a new CRL is published.

There is also a requirement that a single-use EE certificate should have a matching validity period, but I do not think we can distinguish between single-use and sequential-use EE certificates as RP.

rpki-validator/src/main/resources/messages.properties
```diff
@@ -76,6 +76,10 @@ repository.object.maximum.size.error=RPKI object {0} size {1} is greater than th
 repository.object.is.trust.anchor.certificate.error=RPKI object {0} is not a trust anchor
 repository.object.is.trust.anchor.certificate.passed=RPKI object {0} is a trust anchor
 
+manifest.crl.next.update.time.matches.error=Manifest next update time {0} does not match CRL next update time {1}
```
wibisono, line 79: Do we need to indicate that match here means within 60 seconds?

wibisono (approved):

rpki-validator/src/main/resources/messages.properties
erikrozendaal: I'm not sure it is very valuable and in the future this time will be reduced (maybe to 0).

PR #306, opened 2020-12-16
Validator is now always ready after initial validation on startup or re-start

### LINKED ISSUE #284: Validator can become "not ready" a large part of the time
When a validator is updating a trust anchor, it indicates to rtr servers that it is "not ready".

When a rtr-server is running it requests the validator for the validated objects. When the objects (effectively: it samples for them) the rtr server can get a non-ready response multiple times in a row.

This increases the time to convergence.

```
$ while true; do; curl http://localhost:9176/api/objects/validated 2>/dev/null | jq '.data.ready'; sleep 5; done
❯ while true; do; curl http://alias.student.utwente.nl:9176/api/objects/validated 2>/dev/null | jq '.data.ready'; sleep 5; done
false
false
false
false
false
false
false
false
false
false
false
true
true
true
true
false
...
```
This likely is related to #283.

@lolepezy and @erikrozendaal could you take a look?

erikrozendaal:
Fixes #284.

On startup the validator will set all trust anchors and repositories to pending and then run validation. This finishes in about 1-2 minutes, obviously depending on network speed, computer speed, etc. After this the validator is ready for RTR requests.


rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
```diff
@@ -152,12 +149,12 @@ public void validateRrdpRpkiRepository(long rpkiRepositoryId) {
...
 
         boolean triggerCaTreeAfter = false;
-        boolean changedAtLeastOneObject = false;
         try {
             final String uri = rpkiRepository.getRrdpNotifyUri();
             if (isRrdpUri(uri)) {
-                changedAtLeastOneObject = rrdpService.storeRepository(rpkiRepository, validationRun);
+                boolean changedAtLeastOneObject = rrdpService.storeRepository(rpkiRepository, validationRun);
 
+                triggerCaTreeAfter = changedAtLeastOneObject || rpkiRepository.isPending();
```
lolepezy, line 157: Can you clarify on this one? Why do we run tree valdiation in case of `rpkiRepository.isPending()`?

lolepezy (approved):

rpki-validator/src/main/java/net/ripe/rpki/validator3/domain/validation/RpkiRepositoryValidationService.java
erikrozendaal: Certificate tree validation skips any objects from a pending repository, so the output is different when a repository is pending. So when the repository switches to failed or downloaded the certificate tree validation needs to run again.

Before this PR this was usually not a problem since all pending repositories would be new and at least one object would be downloaded. But now that repositories become pending on startup it is often the case that no changes are detected since the repository may have been downloaded (e.g. when the server is restarted there is almost no time between the last download and startup).

wibisono (approved):

ties (approved):

PR #312, opened 2021-02-05
draft: Initial EOL warning

ties:
Initial version of the end-of-life warning on the front page.

Final text will be updated on monday, but we could test the RC over the weekend.


ui-validator/pom.xml
```diff
@@ -43,7 +43,7 @@
                 </goals>
                 <phase>generate-resources</phase>
                 <configuration>
-                  <arguments>install</arguments>
+                  <arguments>install --legacy-peer-deps</arguments>
```
lolepezy, line 46: What is it? I don't know NPM well enough to judge.

lolepezy (approved):
Do we need this huge
ui-validator/package-lock.json
?

wibisono (approved):

ui-validator/pom.xml
ties: That is what was needed to resolve the dependencies on Big Sur. We use legacy packages and something was wrong. This was the easy out, and the `package-lock.json` will make sure we install the same versions every time.

ties:
```
Causes npm to completely ignore peerDependencies when building a package tree, as in npm versions 3 through 6.

If a package cannot be installed because of overly strict peerDependencies that collide, it provides a way to move forward resolving the situation.
```

And yeah, our build is node6 based...

PR #318, opened 2021-04-07
Remove xstream as transitive dependency

ties:
As a software engineer I want to exclude xstream from the transitive dependencies so I do not get spurious alerts about vulnerabilities in a unused transitive dependency.


lolepezy (approved):
Perfetto!

PR #324, opened 2021-06-11
Spring Boot 2.5.1

ties:
1 out of many projects I'm bumping...


ties:
Wish those projects were on a ecosystem with modern tooling btw.

